PR #11, opened 2019-06-30
Cache generated snark files and node_modules separately

weijiekoh:
This PR introduces changes to the CircleCI workflow to make it take less time.

Broke `ci.sh` into separate files to build snarks and run tests respectively.

Cached the contents of the `snark/` directory so it doesn't have to be regenerated with every workflow.


kobigurk:
Amazing!

PR #13, opened 2019-07-04
In-memory storage wrapper as a RocksDB alternative

weijiekoh:
`sbmtjs/src/storage/memory/index.js` implements a simple in-memory key-value store so that we don't have to use RocksDB if we don't want to, e.g. if we want to work with the merkle tree in-browser.

This PR isn't completely ready yet because `semaphorejs/test/contracts/semaphore.js` can't do this:

```
const MemStorage = require('zkp-sbmtjs/src/storage/memory');
```

Instead, it does this:

```
const MemStorage = require('../../../sbmtjs/src/storage/memory')
```

@kobigurk I think this requires an NPM publish step first. Do you think this is a good course of action?

Thanks!


weijiekoh:
Deprecated in favour of https://github.com/kobigurk/semaphore/pull/18

PR #19, opened 2019-07-26
feat: optimizations and bug fixes

kobigurk:
* Optimizes gas usage by removing the signals tree, replacing with a mapping.
* Fixes nullifier bug by upgrading to latest snarkjs. Adds negative tests for the bug.
* Saves all the leaves.
* Upgrades to Solidity 0.5.10.


kobigurk:
@weijiekoh can you please review?

semaphorejs/contracts/Ownable.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.24;
+pragma solidity ^0.5.0;
```
weijiekoh, line 1: I suggest upgrading to the OZ 2.0 Ownable contract.

I can do this in a separate PR as I'd like to add revert messages too.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v2.3.0/contracts/ownership/Ownable.sol

weijiekoh:
`MerkleTreeTester.sol` needs upgrading to 0.5.x too

weijiekoh:
The cached `verifier.sol` is causing the CI build to fail.

kobigurk:
> The cached `verifier.sol` is causing the CI build to fail.

Ah! That's what it is. Thanks.

kobigurk:
Fixes https://github.com/kobigurk/semaphore/issues/16.

semaphorejs/contracts/Ownable.sol
kobigurk: Sounds good, let's leave it to a separate PR.

PR #21, opened 2019-07-30
Revert reason strings

weijiekoh:
This PR adds revert reason strings to the `require` statements in the contracts.

It also upgrades the snarkjs git commit hash in `package.json` to the latest version which now contains a revert reason in the scalar field check in `verifier.sol`.

The Semaphore contract tests now also check for said revert reasons. It now uses `preBroadcastRequire` instead of `preBroadcastCheck`, so that the contract reverts with the specific revert reason. However, `preBroadcastCheck` should be kept in the contract a a convenience view function for operators to verify that their inputs to `broadcastSignal` are correct before they spend gas on a transaction.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,14 +19,14 @@ jobs:
       - restore_cache:
           name: restore-npm-cache
           keys:
-            - v1.6-dependencies-{{ checksum "package-lock.json" }}
+            - v1.7-dependencies-{{ checksum "package-lock.json" }}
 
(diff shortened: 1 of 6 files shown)


weijiekoh:
The revert reason string format is as such:

```
ContractName: description of the error
```

The total length of each string should be at most 63 characters. A 64-character revert reason string costs 10 more gas than a 63-character one.

kobigurk (approved):
Looks great!

PR #22, opened 2019-07-31
Remove broadcaster check in Semaphore.sol and the circuit

weijiekoh:
This PR removes the following from Semaphore.sol in `broadcastSignal()`:

```
        address broadcaster = address(input[4]);
        require(broadcaster == msg.sender);
```

This logic is left-over from when Semaphore did gas refunds, and prevents someone from frontrunning the contract to grab the gas (or the fee). But that isn't necessary as the snark verification step will still ensure that it's been hashed to create `msg`, and that the signature is correct, so there's still no way for anyone to cut out the relayer from their profits.

Additionally, applications that use Semaphore, like a mixer, can set the broadcaster's address to something other than their own contract address, like a [burn relay registry](https://ethresear.ch/t/burn-relay-registry-decentralized-transaction-abstraction-on-layer-2/5820).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,14 +19,14 @@ jobs:
       - restore_cache:
           name: restore-npm-cache
           keys:
-            - v1.7-dependencies-{{ checksum "package-lock.json" }}
+            - v1.8-dependencies-{{ checksum "package-lock.json" }}
 
       - run: npm install
 
       - save_cache:
           paths:
             - node_modules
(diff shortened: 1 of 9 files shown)


weijiekoh:
Removed the broadcaster from the circuit's public inputs too. A future PR will prevent frontrunning `broadcastSignal` using a flag-enabled `onlyOwner` guard.

kobigurk (approved):
LGTM!

PR #23, opened 2019-07-31
fix+feat: fixes some aliasing bugs, adds an optional permissioning to broadcast

kobigurk:
* Aliasing bugs in the public key and the nullifiers hash fixed.
* Adds a default permissioning mode to broadcast, which can be disabled by the owner.
* Changes signal hasher to Keccak256.


kobigurk:
@weijiekoh please review!

semaphorejs/contracts/Semaphore.sol
```diff
@@ -38,17 +39,17 @@ contract Semaphore is Verifier, MultipleMerkleTree, Ownable {
 
     event SignalBroadcast(bytes signal, uint256 nullifiers_hash, uint256 external_nullifier);
 
+    modifier onlyOwnerIfPermissioned() {
+        require(!is_broadcast_permissioned || isOwner(), "Semaphore: Broadcast permission denied");
```
weijiekoh:
```suggestion
        require(!is_broadcast_permissioned || isOwner(), "Semaphore: broadcast permission denied");
```

weijiekoh: I'd use lowercase here to keep it consistent with the other revert strings

semaphorejs/contracts/Semaphore.sol
```diff
@@ -136,4 +134,8 @@ contract Semaphore is Verifier, MultipleMerkleTree, Ownable {
     function setExternalNullifier(uint256 new_external_nullifier) public onlyOwner {
       external_nullifier = new_external_nullifier;
     }
+
+    function disablePermissioning() public onlyOwner {
```
weijiekoh: Should we have `setPermissioning(bool _newPermission)` instead? In case there is a need to re-enable permissioning after it's disabled?

semaphorejs/test/contracts/semaphore.js
```diff
@@ -221,6 +229,28 @@ contract('Semaphore', function (accounts) {
         assert.equal(failed, true);
         assert.equal(reason, 'verifier-gte-snark-scalar-field');
 
+        await semaphore.transferOwnership(accounts[1]);
```
weijiekoh, line 371: We probably should split this up into more `it()` blocks

weijiekoh (changes requested):

semaphorejs/contracts/Semaphore.sol
```diff
@@ -89,9 +90,6 @@ contract Semaphore is Verifier, MultipleMerkleTree, Ownable {
         uint[4] memory input,
         uint256 signal_hash
     ) public {
-        // Note that we only verify the broadcaster's address (input[4]) in the
```
weijiekoh, line 92: Oops! Thanks for spotting this!

PR #24, opened 2019-08-07
Fixes the permissioning test due to a typo in run_ganache.sh

weijiekoh:
There was a typo in `run_ganache.sh` which only set up one account for `ganache-cli`. This caused the permissioning test in `semaphore.js` to fail as it uses `accounts[1]`. This PR fixes that typo.


weijiekoh:
Closing as I think we should consolidate this fix into #25

PR #25, opened 2019-08-07
Preparation for an audit

weijiekoh:
This PR mainly adds documentation comments to the code.

It also changes `Semaphore.roots()` to `root`, which is a more accurate name.

It also fixes some bugs:

- `run_ganache.sh` where a typo caused only one account to be available. 
- the websnark dependency in `package.json`


kobigurk:
Amazing, thanks a lot for that!
Did you notice that the Merkle Tree now stores the leaves? If you still store it in the mixer, you might not have to anymore.

weijiekoh:
Awesome, thanks for the heads up!

kobigurk (approved):

PR #27, opened 2019-08-24
feat: adds id trapdoor and upgrades deps versions

kobigurk:


kobigurk:
@weijiekoh Could you take a look?

weijiekoh:
All looks great!

Clients should provide a 31-byte random value as the identity trapdoor, correct?

kobigurk:
Exactly. Thanks!

PR #32, opened 2019-10-26
To add a mapping which associates signal indices to external nullifiers

weijiekoh:
This adds a mapping `signal_index_to_external_nullifier` to the Semaphore contract which lets clients look up the external nullifier per signal index.

This enables the "Signal History" table at the bottom of this demo UI, which is just a static website:

https://weijiekoh.github.io/semaphore-ui/


weijiekoh:
Closing this as it's outdated

PR #61, opened 2021-10-10
version 2.0

akinovak:
* Switched to poseidon hash
* Identity commitment is calculated as hash(hash(trapdoor, nullifier)) for rln compatibility 
* Incremental merkle tree with 5 nodes
* Omitted signature verification for the sake of saving time


contracts/Hasher.sol
```diff
@@ -0,0 +1,51 @@
+//SPDX-License-Identifier: Unlicense
```
weijiekoh: I recall that the original version of Semaphore is GPL-3.0 licensed. The license for this updated version may be different but I'm not sure. Was there a conversation with Thore/Barry about this?

circuits/semaphore-base.circom
```diff
@@ -0,0 +1,92 @@
...
+    signal private input path_elements[n_levels][LEAVES_PER_PATH_LEVEL];
+
+    signal output root;
+    signal output nullifierHash;
+
+    component secret = CalculateSecret();
+    secret.identity_nullifier <== identity_nullifier;
+    secret.identity_trapdoor <== identity_trapdoor;
+
+    signal secret_hash <-- secret.out;
```
weijiekoh:
```suggestion
    signal secret_hash;
    secret_hash <== secret.out;
```

A best practice in circom is to avoid `<--` whenever possible.

package.json
```diff
@@ -1,17 +1,36 @@
 {
   "name": "semaphore",
-  "version": "0.0.1",
-  "description": "Solidity contracts, zk-SNARk circuits, and test scripts for the Semaphore zero-knowledge signalling gadget",
-  "main": "build/index.js",
+  "version": "1.0.0",
+  "description": "This is the original version of semaphore where Poseidon hash is used for all proofs and additional security with signature proof is removed",
```
weijiekoh:
```suggestion
  "description": "Solidity contracts, zk-SNARK circuits, and test scripts for the Semaphore zero-knowledge signalling gadget. This version uses the Poseidon hash and does not use EdDSA signatures.",
```

package.json
```diff
@@ -1,17 +1,36 @@
...
-    "build": "lerna run build"
+    "test": "ts-node test/verify.ts"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/appliedzkp/semaphore.git.git"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
weijiekoh: The original version of Semaphore used GPL 3.0. We should chat with the team to discuss which version to use.

circuits/semaphore-base.circom
```diff
@@ -0,0 +1,92 @@
...
+    hasher.inputs[1] <== identity_trapdoor;
+    out <== hasher.out;
+}
+
+template CalculateIdentityCommitment() {
+    signal input secret_hash;
+
+    signal output out;
+
+    component hasher = Poseidon(1);
```
weijiekoh, line 23: Which JS function does `poseidon([x])` (only 1 input)? I'm under the impression that circomlib's Poseidon hash requires at least 2 inputs. I might be wrong though.

weijiekoh (commented):
Thanks for this! I've added some initial review comments.

circuits/semaphore-base.circom
akinovak: I didn't face any problems with using poseidon function from circomlib.

https://github.com/iden3/circomlib/blob/39a01fd913bce3c9b2b77f021236a65cee58166b/src/poseidon.js#L22

package.json
```diff
@@ -1,17 +1,36 @@
...
+  "bugs": {
+    "url": "https://github.com/appliedzkp/semaphore.git/issues"
+  },
+  "homepage": "https://github.com/appliedzkp/semaphore.git#readme",
+  "dependencies": {
+    "circomlib": "^0.5.3",
+    "libsemaphore": "github:appliedzkp/libsemaphore",
```
adust09: `libsemaphore` not found :(
 I run `npm install` on this branch, but got an error because the package was not found. 

Have you changed to a private repository or something?
(@akinovak The link you mented to me was also 404.)

akinovak: Hmm, that strange, let me check that asap

akinovak: I'm not getting that error.
 
Can you try with `npm install https://github.com/appliedzkp/libsemaphore.git`.

If it's still not working, let's move this chat somewhere else :)

akinovak: Hey, pls do `npm i` on semaphore again :)

adust09: oh,I try `npm i`,successfly work!
I'm sorry to take up so much of your time.

akinovak: No worries, don't hesitate to ask more questions :)

scripts/build-circuits.sh
```diff
@@ -0,0 +1,29 @@
...
+    wget https://hermez.s3-eu-west-1.amazonaws.com/powersOfTau28_hez_final_16.ptau
+fi
+
+npx snarkjs zkey new semaphore.r1cs powersOfTau28_hez_final_16.ptau semaphore_0000.zkey
+
+npx snarkjs zkey contribute semaphore_0000.zkey semaphore_final.zkey
+
+npx snarkjs zkey export verificationkey semaphore_final.zkey verification_key.json
+
+snarkjs zkey export solidityverifier semaphore_final.zkey verifier.sol
```
adust09: running ".scripts/build-circuits.sh", it generates "Verifier.sol", but this solidity version is "^0.6.11", don't we need to change it to "^0.8.0"?

akinovak: Yep, verifier is automatically generated like that, just change to version you need when using :)

adust09: I see, as long as I change the version by myself, the compiled program will work fine, right?

akinovak: that's right :)

circuits/semaphore-base.circom
```diff
@@ -0,0 +1,90 @@
...
+        inclusionProof.path_elements[i][j] <== path_elements[i][j];
+      }
+      inclusionProof.path_index[i] <== identity_path_index[i];
+    }
+
+    root <== inclusionProof.root;
+
+    // Dummy square to prevent tampering signalHash
+    signal signal_hash_squared;
+    signal_hash_squared <== signal_hash * signal_hash;
```
recmo, line 87: What kind of tampering and how does this prevent it?

Does Circom not remove this as dead code since `signal_hash_squared` is unused?

recmo: To answer my first question: `signal_hash` would otherwise be unused and we need the proof verification to fail if this public input changes.

akinovak: So my argument for this is that since signal is not used in circuit and let's say you are using some realyer, it can temper the signal but proof can still be valid. 

But yes, your answer is good!

recmo: Here's a relevant snip from the circom doc [source](https://docs.circom.io/2.-circom-fundamentals/circuit-structure#compiler-messages):

> If a signal is not used in any constraint, an *error* message will be generated. Moreover, if such signal is an input signal `s` , then the compiler would suggest adding a constraint of the form `x * 0 === 0`. 

They recommend a simpler constraint, but it looks like adding a dummy constraint is the recommended way to solve the issue of unused public inputs. So no risk of Circom optimizing it away.

PR #79, opened 2022-02-04
Feat/modular contracts

cedoor:


akinovak:
Everything seems fine, just not sure if nullifier should be decoupled

PR #91, opened 2022-03-09
Fix/group id value

### LINKED ISSUE #90: Missing checks to ensure zk proof inputs are less than SNARK_SCALAR_FIELD
**Description**
The SemaphoreGroup.sol contract uses the "groupId" as the external nullifier. Therefore, the groupId will be provided as a public input when calling Verifier.verifyProof. When a user creates a new poll via SemaphoreVoting.createPoll, the user can enter any uint256 value as the poll id, which will then be the group Id. The issue arises when the user inputs a value for pollId that is greater than the SNARK_SCALAR_FIELD constant. Then, the call to verify any proofs for this poll will fail because the Verifier checks:
`require(input[i] < snark_scalar_field, "verifier-gte-snark-scalar-field");`
https://github.com/appliedzkp/semaphore/blob/main/contracts/base/Verifier.sol#L285
for each public input.

Since both the SemaphoreVoting.sol and SemaphoreWhistleblowing.sol contracts use the groupId as external nullifiers, this issue is present in both.

**Possible Fix**
Add a constraint in  the SemaphoreGroup.sol contract to check that the groupId is less than the SNARK_SCALAR_FIELD. This will protect any extensions that use the group  id as an external nullifier.

cedoor:
## Pull request checklist

Please check if your PR fulfills the following requirements:

- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`yarn compile`) was run locally and any changes were pushed
- [x] Lint (`yarn lint`) has passed locally and any fixes were made for failures

## Pull request type

Please check the type of change your PR introduces:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

- [ ] Yes
- [x] No

### DIFF
--- a/contracts/base/SemaphoreGroups.sol
+++ b/contracts/base/SemaphoreGroups.sol
@@ -29,6 +29,7 @@ abstract contract SemaphoreGroups is Context, ISemaphoreGroups {
     uint8 depth,
     uint256 zeroValue
   ) internal virtual {
+    require(groupId < SNARK_SCALAR_FIELD, "SemaphoreGroups: group id must be < SNARK_SCALAR_FIELD");
     require(getDepth(groupId) == 0, "SemaphoreGroups: group already exists");
 
     groups[groupId].init(depth, zeroValue);
--- a/contracts/package.json
+++ b/contracts/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@appliedzkp/semaphore-contracts",
-  "version": "0.6.0",
+  "version": "0.6.1",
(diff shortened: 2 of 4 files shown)


Chavrzp28 (commented):
Configuracion de world id

PR #96, opened 2022-03-23
Cleanup Verifier.sol and make error handling consistent

recmo:
## Pull request checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [ ] Build (`yarn compile`) was run locally and any changes were pushed
- [ ] Lint (`yarn lint`) has passed locally and any fixes were made for failures

## Pull request type

Please check the type of change your PR introduces:

- [x] Bugfix
- [ ] Feature
- [x] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

Currently when provided with an invalid proof, the verifier will either:

* return `false`,
* execute the `invalid()` opcode,
* revert with one of a variety of error messages, like
  * `verifier-gte-snark-scalar-field`
  * `pairing-add-failed`,
  * etc..

This makes handling invalid proofs hard. Returning `false` is especially dangerous because it puts the burden on the caller.

## What is the new behavior?

When provided with an invalid proof, the verifier will:

* revert with `InvalidProof()`

In the unlikely event that the caller desires different behaviour, the caller can use a Solidity try/catch statement. 

- [x] Yes
- [ ] No

The revert error message for an invalid proof changes.

### DIFF
--- a/contracts/base/SemaphoreCore.sol
+++ b/contracts/base/SemaphoreCore.sol
@@ -14,7 +14,7 @@ contract SemaphoreCore is ISemaphoreCore {
   /// It is used to prevent double-signaling.
   mapping(uint256 => bool) internal nullifierHashes;
 
-  /// @dev Asserts that no nullifier already exists and that the zero-knowledge proof is valid.
+  /// @dev Asserts that no nullifier already exists and if the zero-knowledge proof is valid.
   /// Otherwise it reverts.
   /// @param signal: Semaphore signal.
   /// @param root: Root of the Merkle tree.
--- a/contracts/base/SemaphoreCore.sol
+++ b/contracts/base/SemaphoreCore.sol
@@ -14,7 +14,7 @@ contract SemaphoreCore is ISemaphoreCore {
   /// It is used to prevent double-signaling.
   mapping(uint256 => bool) internal nullifierHashes;
 
-  /// @dev Returns true if no nullifier already exists and if the zero-knowledge proof is valid.
+  /// @dev Asserts that no nullifier already exists and that the zero-knowledge proof is valid.
   /// Otherwise it reverts.
   /// @param signal: Semaphore signal.
   /// @param root: Root of the Merkle tree.
@@ -22,14 +22,14 @@ contract SemaphoreCore is ISemaphoreCore {
   /// @param externalNullifier: External nullifier.
   /// @param proof: Zero-knowledge proof.
   /// @param verifier: Verifier address.
-  function _isValidProof(
+  function _verifyProof(
     bytes32 signal,
     uint256 root,
     uint256 nullifierHash,
     uint256 externalNullifier,
     uint256[8] calldata proof,
     IVerifier verifier
-  ) internal view returns (bool) {
+  ) internal view {
     require(!nullifierHashes[nullifierHash], "SemaphoreCore: you cannot use the same nullifier twice");
 
     uint256 signalHash = _hashSignal(signal);
@@ -40,8 +40,6 @@ contract SemaphoreCore is ISemaphoreCore {
       [proof[6], proof[7]],
       [root, nullifierHash, signalHash, externalNullifier]
     );
-
-    return true; // TODO: This is redundant
   }
 
   /// @dev Stores the nullifier hash to prevent double-signaling.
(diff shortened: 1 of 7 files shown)


recmo:
My next move would be to remove the return value from `_isValidProof`, but at that point the name should also change. I want to check in and get maintainers opinion first.

Personally I prefer to revert on invalid proofs. This prevents errors where the result is accidentally ignored. (See for example also `safeTransfer` and why it exists)  I don't think there are cases where callers would want any thing else, but if so they can always intercept the using try/catch.

kobigurk (commented):
Are you sure EIP196 verifies input is in the correct range?

Specifically:

image

cedoor (commented):
Hi! After the next Semaphore ceremony we will have more verifiers (> 10). Since this code is regenerated every time from SnarkJS templates I think it is better to create a new [`verifier_groth16.sol.ejs`](https://github.com/iden3/snarkjs/blob/master/templates/verifier_groth16.sol.ejs) template with your changes.

recmo:
@kobigurk  Looks like you are partially right: field elements are verified, but scalars are not considered field elements.

Screen Shot 2022-03-24 at 09 19 21

kobigurk:
> @kobigurk Looks like you are partially right: field elements are verified, but scalars are not considered field elements.
> 
> Screen Shot 2022-03-24 at 09 19 21

Ah yes, exactly the distinction I was looking for. Super dangerous :)

recmo:
Ok, so let's be conservative here and reject proofs where the input is not reduced. I'll add a commit.

kobigurk:
> Ok, so let's be conservative here and reject proofs where the input is not reduced. I'll add a commit.

Thanks! Btw, it's not only conservative, it's a direct attack vector - https://github.com/a16z/zkp-merkle-airdrop-contracts/pull/2

recmo:
I made all the functions in `Pairing`  strict on the input. Note that before `Pairing.negate(..)` would accept an unreduced `Y` coordinate, so there was malleability in `Proof.A`.

I don't know if this removes all proof malleability for people other than the prover. (i.e. given a valid proof but not the private inputs, can I compute a different also valid proof)

recmo:
@cedoor I suspected there was some template involved, despite the usual warning comment missing. Thanks for the link, I will submit a PR there too.

I think we can still merge it here, the only thing that changes with a new ceremony is the contents of the `verifyingKey()` function, so that should be easy to update. The only other thing that can possibly change (assuming same proof system) is the number of inputs, which is also trivially changed in `verifyProof()`.

kobigurk:
> I made all the functions in `Pairing` strict on the input. Note that before `Pairing.negate(..)` would accept an unreduced `Y` coordinate, so there was malleability in `Proof.A`.
> 
> I don't know if this removes all proof malleability for people other than the prover. (i.e. given a valid proof but not the private inputs, can I compute a different also valid proof)

It removes "public input" malleability, which is the dangerous part. Proofs can still be modified, because Groth16 is randomizable.

contracts/base/SemaphoreCore.sol
```diff
@@ -34,13 +34,14 @@ contract SemaphoreCore is ISemaphoreCore {
...
-        [root, nullifierHash, signalHash, externalNullifier]
-      );
+    verifier.verifyProof(
+      [proof[0], proof[1]],
+      [[proof[2], proof[3]], [proof[4], proof[5]]],
+      [proof[6], proof[7]],
+      [root, nullifierHash, signalHash, externalNullifier]
+    );
+
+    return true; // TODO: This is redundant
```
cedoor: Can't we remove this line? Updating the extension contracts should be sufficient to make this change compatible

recmo: Happy to do so! I left it to limit the scope of changes.

cedoor (approved):

contracts/base/SemaphoreCore.sol
cedoor: Seems you need to run `yarn prettier:fix` to pass tests.

PR #154, opened 2022-11-08
Feat/build script

taskooh:
## How to build circom
- install circom 
- run following command:
```
yarn build:circom
```


cedoor (commented):
Hey @taskooh, thank you very much for this PR!

It would be great to have a JS script, so that it can be used in different environments. Do you think it is possible to convert your bash script?

PR #191, opened 2022-12-15
Semaphore CLI

### LINKED ISSUE #182: Semaphore CLI
**Describe the package you'd like**

A CLI to initialize Semaphore projects with 1 command: `semaphore init`, with options to choose your favorite testing framework ([Foundry](https://github.com/foundry-rs/foundry) or [Hardhat](https://hardhat.org/)), language (JavaScript or TypeScript) etc.

The same CLI might be extended with other commands then, like commands to query on-chain data (connected to the subgraph):

- `semaphore get-groups`
- `semaphore get-group <group-id>`
- `semaphore get-group <group-id> --members --signals`

Foundry can also further increase the adoption of Semaphore, since it is increasingly being used.

cedoor:
## Description

Setting up a project, although not particularly complex, can be a lengthy process for some people. A command line to initialize a Semaphore project can reduce the set-up time from a few minutes to a few seconds. The command line can also be useful for obtaining the data of groups deployed on the networks supported by Semaphore.

This PR adds a new `@semaphore-protocol/cli` package which can be used to initialize Semaphore projects and get groups' data. It also includes a mechanism to dynamically download templates as NPM packages. There is currently only one template (`@semaphore-protocol/cli-template-hardhat`), but new ones can be integrated in the future.

## Related Issue

Closes #182

-   [ ] Yes
-   [x] No


packages/cli/src/index.ts
```diff
@@ -0,0 +1,152 @@
...
+        await download(templateURL, currentDirectory, { extract: true })
+
+        renameSync(`${currentDirectory}/package`, `${currentDirectory}/${projectDirectory}`)
+
+        spinner.stop()
+
+        console.info(` ${logSymbols.success}`, `Your project is ready!\n`)
+        console.info(` Please, install your dependencies by running:\n`)
+        console.info(`   ${chalk.cyan("cd")} ${projectDirectory}`)
+        console.info(`   ${chalk.cyan("npm i")}\n`)
```
sripwoud, line 55: Why not having the semaphore CLI execute those steps too?
I am thinking about `child_process.exec`
And then something like
```
exec(`cd ${projectDirectory}`)
exec('npm i')
```

packages/cli/src/index.ts
```diff
@@ -0,0 +1,152 @@
...
+                    .map((s) => `   ${chalk.cyan(`npm run ${s}`)}`)
+                    .join("\n")}\n`
+            )
+        }
+    })
+
+program
+    .command("get-groups")
+    .description("Get the list of groups from a supported network (Goerli or Arbitrum).")
+    .option("-n, --network <network-name>", "Supported Ethereum network.", "goerli")
```
sripwoud: Maybe precise here that for arbitrum the possible arg is `arbitrum` and not `Arbitrum`

packages/cli/src/index.ts
```diff
@@ -0,0 +1,152 @@
...
+    .action(async ({ network }) => {
+        const subgraph = new Subgraph(network)
```
sripwoud, line 80: Check that `network` is supported and return a nice error by the CLI instead of throwing?
```
node ../semaphore/packages/cli/dist/index.js get-groups -n Arbitrum
file:///home/r1oga/dev/semaphore/packages/subgraph/dist/index.mjs:88
            throw new TypeError("Network '".concat(network, "' is not supported"));
```

Or better, use inquirer with choices prompt
https://github.com/SBoudrias/Inquirer.js#prompt-types
that way we control what the user can select

packages/cli/src/index.ts
```diff
@@ -0,0 +1,152 @@
...
+
+program
+    .command("get-group")
+    .description("Get the data of a group from a supported network (Goerli or Arbitrum).")
+    .argument("<group-id>", "Identifier of the group.")
+    .option("-n, --network <network-name>", "Supported Ethereum network.", "goerli")
+    .option("--members", "Show group members.")
+    .option("--signals", "Show group signals.")
+    .action(async (groupId, { network, members, signals }) => {
+        const subgraph = new Subgraph(network)
```
sripwoud, line 118: same check that network is valid?

For `groupId`, I am not sure a select makes sense... too many values to choose from

sripwoud (commented):
I could successfully run all the operations of the CLI :ok_hand: 

I suggest looking at https://github.com/SBoudrias/Inquirer.js
That way the cli is more interactive and more "fool proof" because you control better what the user can choose

cedoor:
> I could successfully run all the operations of the CLI ok_hand
> 
> I suggest looking at https://github.com/SBoudrias/Inquirer.js That way the cli is more interactive and more "fool proof" because you control better what the user can choose

Yep, it looks cool. We can definitely use Inquirer as soon as we'll have more templates in the `init` command! I'll create a separate issue. Thanks 🙏🏽

packages/cli/src/index.ts
cedoor: Definitely something I'd like to have 👍🏽  It would also need a check to understand which package manager the user is using though.

sripwoud: Isn't `npm` always bundled with a node installation? Can't we assume `npm` commands will always work?
But yeah the drawback is that you will install with npm while the user may have preferred something else.

packages/cli/src/index.ts
cedoor: Thanks, I'll add a check. I tend to prefer both approaches together. For example: a user can use the optional `-n` option OR the prompt in this case.

packages/cli/src/index.ts
cedoor: Too many values yep, there could be thousands

PR #247, opened 2023-02-08
Integrate Inquirer.js in create command in the CLI

### LINKED ISSUE #192: Integrate InquirerJS

vplasencia:
## Description

Integrate Inquirer.js in `create` command in the CLI so that when running `semaphore create` without the project name, a question appears to get the project name. 

## Related Issue

Closes #192 

-   [ ] Yes
-   [x] No

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -7,6 +7,7 @@ import { existsSync, readFileSync, renameSync } from "fs"
 import logSymbols from "log-symbols"
 import { dirname } from "path"
 import { fileURLToPath } from "url"
+import inquirer from "inquirer"
 import checkLatestVersion from "./checkLatestVersion.js"
 import Spinner from "./spinner.js"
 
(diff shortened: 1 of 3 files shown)


packages/cli/src/index.ts
```diff
@@ -30,15 +31,29 @@ program
...
+            const answers = await inquirer.prompt({
+                name: "project_name",
+                type: "input",
+                message: "Enter your project name:",
+                default: "my-app"
+            })
+            projectDir = answers.project_name
+        } else {
+            projectDir = projectDirectory
+        }
```
cedoor:
```suggestion
        if (!projectDirectory) {
            const answers = await inquirer.prompt({
                name: "project_name",
                type: "input",
                message: "Enter your project name:",
                default: "my-app"
            })
            
            projectDirectory = answers.project_name
        }
```

`projectDirectory` can directly be used, since it's not a constant. But it's necessary to rename the `projectDir` variable below.

cedoor (changes requested):

cedoor (approved):

PR #250, opened 2023-02-15
Add inquirer in get-groups and get-group commands

vplasencia:
## Description

Add inquirer in get-groups and get-group commands and improve the developer experience in the CLI.

-   [ ] Yes
-   [x] No

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -38,13 +38,13 @@ program
     .allowExcessArguments(false)
     .action(async (projectDirectory) => {
         if (!projectDirectory) {
-            const answer = await inquirer.prompt({
-                name: "project_name",
+            const { projectName } = await inquirer.prompt({
+                name: "projectName",
                 type: "input",
                 message: "What is your project name?",
                 default: "my-app"
             })
-            projectDirectory = answer.project_name
+            projectDirectory = projectName
         }
 
         const currentDirectory = process.cwd()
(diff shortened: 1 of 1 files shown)


packages/cli/src/index.ts
```diff
@@ -123,13 +136,74 @@ program
...
+                })
+                groupId = selectedGroupId
+            } catch (error) {
+                spinnerGroups.stop()
+                console.info(`\n ${logSymbols.error}`, "error: unexpected error with the Semaphore subgraph")
+                return
+            }
+        }
+
+        if (!members && !signals) {
```
cedoor, line 186: I was thinking that it might make sense to not manage those options with Inquirer when the user provides all the required parameters, because they are optional and booleans and it wouldn't be possible to run the `semaphore get-group 1 -n goerli` without user interaction (e.g in CI env).

What do you think?

packages/cli/src/index.ts
```diff
@@ -123,13 +136,74 @@ program
...
+    .option("-m, --members", "Show group members.")
+    .option("-s, --signals", "Show group signals.")
     .allowExcessArguments(false)
     .action(async (groupId, { network, members, signals }) => {
-        if (!["goerli", "arbitrum"].includes(network)) {
+        if (!network) {
+            const { selectedNetwork } = await inquirer.prompt({
+                name: "selectedNetwork",
+                type: "list",
+                message: "What is the network?",
```
cedoor: These prompt are created to show a list of options. I'd update the massage with something like "Select one of our supported networks:", like [`cz`](https://github.com/commitizen/cz-cli). "What is the network?" makes more sense when there are not options but only open answers.

"What is the group id?" -> "Select one of the following existing group ids" or something like that

How does it sound?

cedoor (commented):
Inquirer is cool! Great work 🔥

packages/cli/src/index.ts
vplasencia: Yes, I agree. What about using `network` as an argument in `get-groups` and `get-group` commands and managing only arguments with Inquirer?

packages/cli/src/index.ts
vplasencia: Sounds good.

packages/cli/src/index.ts
cedoor: What do you mean by "managing only arguments"?

vplasencia: After applying your suggestion and if `network` is used as an argument, Inquirer will be used for arguments only.

PR #251, opened 2023-02-15
Subgraph library improvements

saleel:
## Description

- Add query filters to subgraph library
- Add option to specify subgraph URL in subgraph library

## Related Issue

#242 
#243 

-   [ ] Yes
-   [x] No

### DIFF
--- a/packages/subgraph/src/subgraph.ts
+++ b/packages/subgraph/src/subgraph.ts
@@ -3,6 +3,26 @@ import checkParameter from "./checkParameter"
 import getURL from "./getURL"
 import request from "./request"
 import { GroupOptions, Network } from "./types"
+import { jsDateToGraphqlDate } from "./utils"
+
+type GetGroupsResponse = {
+    id: string
+    merkleTree: {
+        root: string
+        depth: number
+        zeroValue: string
+        numberOfLeaves: number
+    }
+    admin: string
+    members: string[]
+    verifiedProofs: {
+        signal: string
+        merkleTreeRoot: string
+        externalNullifier: string
+        nullifierHash: string
+        timestamp: string
+    }[]
+}
 
 export default class Subgraph {
     private _url: string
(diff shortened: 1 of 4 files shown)


packages/subgraph/src/subgraph.ts
```diff
@@ -30,19 +55,40 @@ export default class Subgraph {
      * @param options Options to select the group parameters.
      * @returns List of groups.
      */
-    async getGroups(options: GroupOptions = {}): Promise<any[]> {
+    async getGroups(options: GroupOptions = {}): Promise<GetGroupsResponse[]> {
```
cedoor: Should we add the `GetGroupsResponse` return type to the `getGroup` function too?

packages/subgraph/src/subgraph.ts
```diff
@@ -3,6 +3,26 @@ import checkParameter from "./checkParameter"
...
+    }
+    admin: string
+    members: string[]
+    verifiedProofs: {
+        signal: string
+        merkleTreeRoot: string
+        externalNullifier: string
+        nullifierHash: string
+        timestamp: string
+    }[]
```
cedoor:
```suggestion
    members?: string[]
    verifiedProofs?: {
        signal: string
        merkleTreeRoot: string
        externalNullifier: string
        nullifierHash: string
        timestamp: string
    }[]
```

cedoor (changes requested):

packages/subgraph/src/subgraph.ts
saleel: Ah yes, missed that. Added now.

packages/subgraph/src/subgraph.ts
saleel: I thought of adding Partial<> to the response, but then consuming apps would need to typecast as not null. 
I have added this now anyway.

cedoor (approved):

PR #262, opened 2023-03-04
feat(identity): use sha512 to extract more entropy from message

vimwitch:
## Description

In the current identity scheme the nullifier and trapdoor use very similar pre-images. If either the nullifier or trapdoor is compromised the other value can be determined easily (replace the appended value `identity_nullifier` with `identity_trapdoor`).

Instead of using a keccak256 hash for both identity nullifier and trapdoor we can use a sha512 hash and split the output in half, using the first 256 bits for nullifier, and the second for a trapdoor. We can also bit shift by 3 instead of 8 for slightly more protection.

-   [x] Yes
-   [ ] No


vimwitch:
@cedoor not sure how you feel about including another hashing library, or breaking existing message based identities. Just thought i'd throw this out there

cedoor:
> @cedoor not sure how you feel about including another hashing library, or breaking existing message based identities. Just thought i'd throw this out there

Just read the code. The possibilities to find one identity secret and then its pre-image are very low, however you're right, it can make the identities even more secure.

Unless a user loses one of the identity secrets (in that case they will probably lose both the secrets btw), the simplest attack vector for an attacker should be as follows (let me know if it makes sense to you): calculate the pre-image of the nullifier hash (since the external nullifier is public), i.e. the nullifier identity, and calculate its pre-image again. Then, as you said, they can easily find the identity trapdoor because the pre-image is almost the same, and generate a proof.

Your changes should add a further step, so they would need to calculate 3 pre-images vs 2 pre-images. The third one would be the pre-image of the identity commitment knowing the identity nullifier. Correct?

I think we can merge it, as there shouldn't be prod apps yet.

vimwitch:
> Your changes should add a further step, so they would need to calculate 3 pre-images vs 2 pre-images. The third one would be the pre-image of the identity commitment knowing the identity nullifier. Correct?

Yes, this makes sense. It should add just a bit more security.

cedoor (approved):

PR #364, opened 2023-09-22
SemaphoreViem implemented

### LINKED ISSUE #343: Create lib that uses Viem
[Viem](https://viem.sh) is a lighter alternative to [Ethers](https://ethers.org/). This new lib can be part of the `@semaphore-protocol/data` package so that devs can choose what they prefer to use in their applications.

inaridiy:
## Description
SemaphoreViem is implemented.

## Related Issue

Closes #343

-   [ ] Yes
- [x] No


inaridiy:
Worse, I forgot to write the test. I'll go write one soon.

vplasencia:
Hey @inaridiy! Don't worry, take your time. Thank you so much.

inaridiy:
@vplasencia Accomplished! I have confirmed a successful build and test!

packages/data/src/types/index.ts
```diff
@@ -56,3 +56,9 @@ export type EthersOptions = {
     provider?: "etherscan" | "infura" | "alchemy" | "cloudflare" | "pocket" | "ankr"
     apiKey?: string
 }
+
+export type ViemOptions = {
+    address?: `0x${string}`
```
cedoor, line 68: Nice!

packages/data/src/viem.ts
```diff
@@ -0,0 +1,241 @@
...
+
+    /**
+     * Initializes the Viem object with an viem chain.
+     * @param network Chain object of viem.
+     * @param options Viem options.
+     */
+    constructor(network: Chain = sepolia, options: ViemOptions = {}) {
+        if (options.rpcURL) {
+            checkParameter(options.rpcURL, "rpcURL", "string")
+            if (!options.rpcURL.startsWith("http")) throw new Error(`Invalid rpcURL '${options.rpcURL}'`)
```
cedoor: You could use the URL Constructor for a stronger check.

packages/data/src/viem.ts
```diff
@@ -0,0 +1,241 @@
...
+    private _client: PublicClient
+    private _options: ViemOptions
+    private _contract: GetContractReturnType<typeof SemaphoreABI, PublicClient>
+
+    /**
+     * Initializes the Viem object with an viem chain.
+     * @param network Chain object of viem.
+     * @param options Viem options.
+     */
+    constructor(network: Chain = sepolia, options: ViemOptions = {}) {
```
cedoor, line 17: It would be nice for devs to have a way to pass a valid chain without installing `vite` separately and importing `viem/chains`. The type here can be similar to `EthersNetwork` (i.e. strings), and the constructor could then use the right chain object based on that string. It would also be more consistent with `SemaphoreEthers`.

packages/data/src/viem.ts
```diff
@@ -0,0 +1,241 @@
...
+
+        return members
+    }
+
+    /**
+     * Returns a list of group members.
+     * @param groupId Group id.
+     * @returns Group members.
+     */
+    async getGroupVerifiedProofs(groupId: string) {
```
cedoor: This function should have the return value type.

cedoor (changes requested):

inaridiy:
Thanks for the review. I will fix it now.

packages/data/src/viem.ts
inaridiy: I think the current implementation is better here, given its affinity with the Viem ecosystem.

inaridiy:
I have made some corrections and added some comments

packages/data/src/viem.ts
cedoor: Makes sense. Do you think it makes sense to export those chains' objects in this package too?

inaridiy: It certainly makes sense, and I find it a very nice contrast to SemaphoreEthers.

cedoor:
> I have made some corrections and added some comments

Great, it looks like there's a small error in the API test.

cedoor:
Hey @inaridiy, after releasing Semaphore V4 many conflicts have come out. It may make sense to open another clean PR. Do you still want to work on this?

inaridiy:
@cedoor  Thank you for the follow-up. I apologize for my long inactivity on this PR. Given the conflicts with Semaphore V4 and my inability to dedicate proper time, I agree it's best to close this PR.
I still believe in the value of Viem integration, but it deserves a fresh start. If there's interest in the future, I'd be happy to contribute again or support others taking on this task.
Thank you for your understanding. Please feel free to close this PR. I appreciate the opportunity and will keep an eye out for future ways to contribute to Semaphore.

cedoor:
Hi @inaridiy, thanks for your quick reply and the work you've done here.

I'll close this PR but a gitPOAP will still be awarded to you :)

cedoor:
@inaridiy @gitpoap-bot

PR #452, opened 2023-11-15
fix: typescript type exports

dcposch:
## Description

Fix typescript exports.

This makes it easy to install and `import` from `@semaphore-protocol/*` in a latest-version/default-configuration Typescript project.

## Related Issue

Importing `@semaphore/<x>` in a Node 20 Typescript project currently fails with the following error:

image

This appears to be because Typescript is looking for `package.json` > `"exports"` > `"types"`.

-   [ ] Yes
-   [x] No

### DIFF
--- a/packages/data/package.json
+++ b/packages/data/package.json
@@ -6,7 +6,8 @@
     "main": "dist/index.node.js",
     "exports": {
         "import": "./dist/index.mjs",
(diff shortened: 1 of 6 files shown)


cedoor (approved):

cedoor:
@dcposch Thanks a lot!

PR #455, opened 2023-11-28
Show the latest videos and articles first on the learn page

### LINKED ISSUE #453: Show the latest videos and articles first on the learn page

vplasencia:
## Related Issue

Closes #453 

-   [ ] Yes
-   [x] No


cedoor (commented):
Should we use the timestamp to sort videos by date?

vplasencia:
@cedoor we could also use the date parameter (for articles and videos), but then the date should be correct and not empty. What do you think?

cedoor:
> @cedoor we could also use the date parameter (for articles and videos), but then the date should be correct and not empty. What do you think?

I think that makes sense

cedoor (approved):

PR #470, opened 2023-12-11
Home page exceeds page's dimensions on Chromium-based browsers

### LINKED ISSUE #462: Home page exceeds page's dimensions on Chromium-based browsers

0xjei:
## Description

To solve this, I set the `overflowX` property in the body to `hidden` (to prevent the content from overflowing horizontally into the parent element's box) and added `minWidth` and `maxWidth` instead of `width`, stretching the image to the maximum for the page layout size.

## Related Issue
#462

-   [ ] Yes
-   [x] No

### DIFF
--- a/apps/website/src/app/page.tsx
+++ b/apps/website/src/app/page.tsx
@@ -27,7 +27,8 @@ export default function Home() {
                         alt="Midnight whispers image"
                         src="https://semaphore.cedoor.dev/midnight-whispers.jpg"
                         objectFit="cover"
-                        w="full"
+                        minWidth="100%"
+                        maxWidth="100%"
                         h="full"
                     />
                 </Box>
@@ -202,7 +203,8 @@ export default function Home() {
                         alt="Fluttering shadow image"
                         src="https://semaphore.cedoor.dev/shadow-flutter.jpg"
                         objectFit="cover"
-                        w="full"
+                        minWidth="100%"
+                        maxWidth="100%"
                         h="full"
                     />
                 </Box>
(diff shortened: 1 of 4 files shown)


vplasencia (approved):

apps/website/src/app/page.tsx
```diff
@@ -27,7 +27,8 @@ export default function Home() {
                         alt="Midnight whispers image"
                         src="https://semaphore.cedoor.dev/midnight-whispers.jpg"
                         objectFit="cover"
-                        w="full"
+                        minWidth="100%"
```
cedoor: We could not need this. I've tried it with `overflowX: "hidden"` only and it works. Could you double-check?

0xjei: You mean, the `w="full"`?

cedoor: Sorry, * we could not need this.

`overflowX: "hidden"` could be enough for making it work.

0xjei: No, if you remove the parameter `minWidth="100%"` the images do not extend horizontally and you get something like this

image

cedoor: Umh I'm trying it with `w="full"` and it works

0xjei: Yes, maintaining `w="full"` works since is equivalent to min & max at 100%. Lmk which ones do you prefer

cedoor: Mmh I'd reduce the number of changes in this PR to the bare minimum.

0xjei: Better for consistency across images, see last commit

cedoor: There's still another image to update

cedoor: @0xjei

cedoor (approved):

PR #472, opened 2023-12-11
refactor: switched from single to multi selector for projects categories

0xjei:
## Description

This PR enable the selection of multiple project categories. The projects shown are those that have at least all categories selected.

## Related Issue
See #459 

-   [ ] Yes
-   [x] No

## Other information

[Screencast from 12-12-2023 00:49:15.webm](https://github.com/semaphore-protocol/semaphore/assets/20580910/18b7bb7e-62f6-4157-b2ab-202f9b431d72)

### DIFF
--- a/apps/website/src/components/ProjectsList.tsx
+++ b/apps/website/src/components/ProjectsList.tsx
@@ -68,25 +68,28 @@ export default function ProjectsList(props: any) {
                 </HStack>
             </VStack>
 
-            <HStack spacing="3" flexWrap="wrap">
-                {getProjectCategories(allProjects).map((category) => (
-                    <Button
-                        key={category}
-                        size="sm"
-                        variant={selectedCategories.includes(category) ? "solid" : "outline"}
-                        colorScheme={selectedCategories.includes(category) ? "primary" : "inherit"}
-                        onClick={() => {
-                            const newCategories = selectedCategories.includes(category)
-                                ? selectedCategories.filter((c) => c !== category)
-                                : [...selectedCategories, category]
+            <VStack align="left" spacing="6">
+                <Text fontSize="20">Category</Text>
+                <HStack spacing="3" flexWrap="wrap">
+                    {getProjectCategories(allProjects).map((category) => (
+                        <Button
+                            key={category}
+                            size="sm"
+                            variant={selectedCategories.includes(category) ? "solid" : "outline"}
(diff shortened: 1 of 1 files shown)


apps/website/src/components/ProjectsList.tsx
```diff
@@ -12,32 +12,18 @@ import { chunkArray } from "../utils/chunkArray"
...
 
 export default function ProjectsList(props: any) {
-    const [projects, setProjects] = useState<(typeof allProjects)[]>(chunkArray(allProjects))
+    const [projects] = useState<(typeof allProjects)[]>(chunkArray(allProjects))
     const [index, setIndex] = useState<number>(0)
-    const [selectedCategory, setSelectedCategory] = useState<string | null>(null)
     const [onlyPSE, setOnlyPSE] = useState<boolean | null>(null)
+    const [selectedCategories, setSelectedCategories] = useState<Array<string>>([])
 
-    const filterProjects = useCallback(() => {
```
cedoor, line 20: I think it's better to keep this function here and make it work with multi-categories. There were two different filters: 1 for PSE projects and 1 for tags and they should work together (that's why the logic is a little bit more complex than 1 filter). With your function, you're only filtering by tags/categories (nothing happens if I click on PSE/Community).

cedoor (changes requested):

apps/website/src/components/ProjectsList.tsx
0xjei: You right, fixed in a6fde2fdedb3ad70970e643b4a212364ea77c6fa

apps/website/src/components/ProjectsList.tsx
```diff
@@ -66,23 +68,25 @@ export default function ProjectsList(props: any) {
                 </HStack>
             </VStack>
 
-            <VStack align="left" spacing="6">
-                <Text fontSize="20">Category</Text>
```
cedoor, line 70: I think we should keep this text.

cedoor (changes requested):

apps/website/src/components/ProjectsList.tsx
0xjei: I must have missed it, my bad!

cedoor (approved):
LGTM!

vplasencia (approved):

Issue #352, opened 2023-09-01
Remove Goerli from the supported networks


vplasencia:
Goerli is [deprecated](https://ethereum.org/en/developers/docs/networks/#goerli) and will be replaced by [Holesovice](https://github.com/eth-clients/holesovice).

Goerli needs to be removed from the following packages ([semaphore repository](https://github.com/semaphore-protocol/semaphore)):

- [`@semaphore-protocol/data`](https://github.com/semaphore-protocol/semaphore/blob/main/packages/data)
- [`@semaphore-protocol/cli`](https://github.com/semaphore-protocol/semaphore/blob/main/packages/cli)

And from the following repositories:

- https://github.com/semaphore-protocol/subgraph ([`networks.json`](https://github.com/semaphore-protocol/subgraph/blob/main/networks.json), [`README.md`](https://github.com/semaphore-protocol/subgraph/blob/main/README.md))
- https://github.com/semaphore-protocol/website ([`en - deployed-contracts.md`](https://github.com/semaphore-protocol/website/blob/main/versioned_docs/version-V3/deployed-contracts.md), [`es - deployed-contracts.md`](https://github.com/semaphore-protocol/website/blob/main/i18n/es/docusaurus-plugin-content-docs/version-V3/deployed-contracts.md))

However, it makes sense to start working on this issue before releasing Semaphore v4, and in the meantime inform people that Goerli will not be supported anymore soon.

For now a solution can be to add an asterisk to `Goerli`, `Optimism Goerli` and `Arbitrum Goerli` on the [website](https://semaphore.pse.dev/docs/deployed-contracts), add a note like `The Goerli testnet is deprecated` and link the following websites for more information:

* Goerli: https://ethereum.org/en/developers/docs/networks/#goerli
* Optimism Goerli: https://community.optimism.io/docs/useful-tools/networks/#op-goerli 
* Arbitrum Goerli: https://docs.arbitrum.io/for-devs/concepts/public-chains#arbitrum-goerli

sumitvekariya:
Hey, I would like to contribute to this one, 

only `goerli` should be remove from the supported networks? or `Arbitrum Goerli` & `Optimism Goerli` should also be removed?

cedoor:
> only `goerli` should be remove from the supported networks? or `Arbitrum Goerli` & `Optimism Goerli` should also be removed?

[...]

Issue #329, opened 2023-05-29
Add a function to verify proof without storing any nullifier


cedoor:
Description in #366 (closed as duplicate).

PR #480, opened 2023-12-15
Semaphore V4

### LINKED ISSUE #357: Integrate Semaphore v4 changes into the protocol
**Description**:

Update the Semaphore codebase based on the findings from the analysis and research period described in the [research repository](https://github.com/semaphore-protocol/research/).

cedoor:
## Description

This PR updates the Semaphore codebase based on the findings from the analysis and research period described in the [research repository](https://github.com/semaphore-protocol/research/).

The only supported tree depth is 10 at the moment. Another PR (based on issue https://github.com/semaphore-protocol/semaphore/issues/482) will be created to support a tree depth range from 1 to 32.

This is not the final version of the PR. The tasks that will follow are listed below. However, an initial review is necessary before proceeding.

**ZK circuits**

- The identity trapdoor and identity nullifier have been replaced with a single value, the private key, from which an EdDSA public key is derived. The Poseidon hash of the public key is the new identity commitment.
- Some input and output names in the circuits have been changed.
  - External Nullifier → Scope
  - SignalHash → Message
  - NullifierHash → Nullifier
  - Siblings → Tree Siblings
  - Path Indices → Tree Indices
- The new circuit uses [Anonymous Components](https://docs.circom.io/circom-language/anonymous-components-and-tuples/#anonymous-components) to make the code clearer and more concise (24 lines of code without comments, rather than 90).
- The circuit used to calculate the Merkle tree root is now part of an external package ([`@zk-kit/circuits`](https://github.com/privacy-scaling-explorations/zk-kit/tree/main/packages/circuits) - [`BinaryMerkleRoot`](https://github.com/privacy-scaling-explorations/zk-kit/blob/main/packages/circuits/circom/binary-merkle-root.circom)) and supports dynamic depths.
- [Circomkit](https://github.com/erhant/circomkit) has been used for tests and scripts
- The new circuit will be published on NPM (`@semaphore-protocol/circuit`)

**JS libraries**

* The core Semaphore libraries have been updated to support the new [LeanIMT](https://github.com/privacy-scaling-explorations/zk-kit/blob/main/packages/imt/src/lean-imt/lean-imt.ts) data structure and the new V4 circuits.
* The identity library now supports the generation of public/private keys using EdDSA and Baby Jubjub, as well as signatures.
* The group and proof libraries have been greatly simplified.
* The proof library now automatically downloads `wasm`/`zkey` artifacts on browsers and also on NodeJS.
* JS tests have been moved outside the `src` folder. A new `tests` folder has been created for them.
* `@semaphore-protocol/heyauthn` tests have been updated to support the new core libraries
* `@semaphore-protocol/hardhat` has been updated to support the new contracts

**Solidity contracts**

* The contracts have been updated with the new [LeanIMT](https://github.com/privacy-scaling-explorations/zk-kit/blob/main/packages/imt.sol/contracts/internal/InternalLeanIMT.sol).
* `SemaphoreGroups.sol` has been adapted to the new data structure, which does not require zero nodes and has a static depth.
* `SemaphoreGroups.sol` now also manages access to functions (group admins).
* The verifier has been updated with the latest version of the template provided by [SnarkJS](https://github.com/iden3/snarkjs/blob/master/templates/verifier_groth16.sol.ejs).
* The `VerifiedProof` event now includes all the parameters of the proof, allowing anyone to verify the proof off-chain (https://github.com/semaphore-protocol/semaphore/issues/328).

## Related Issue

Closes #357 

-   [x] Yes
-   [ ] No

### Next tasks

Before merging this PR to the `main` branch it is necessary to work on other tasks, which will be merged to this branch first.

- https://github.com/semaphore-protocol/semaphore/issues/530
- https://github.com/semaphore-protocol/semaphore/issues/531
- https://github.com/semaphore-protocol/semaphore/issues/532
- https://github.com/semaphore-protocol/semaphore/issues/533
- https://github.com/semaphore-protocol/semaphore/issues/486
- https://github.com/semaphore-protocol/semaphore/issues/534
- https://github.com/semaphore-protocol/semaphore/issues/352
- https://github.com/semaphore-protocol/semaphore/issues/482
- https://github.com/semaphore-protocol/semaphore/issues/329
- https://github.com/semaphore-protocol/semaphore/issues/485
- https://github.com/semaphore-protocol/semaphore/issues/521
- https://github.com/semaphore-protocol/semaphore/issues/522
- https://github.com/semaphore-protocol/semaphore/issues/524
- https://github.com/semaphore-protocol/semaphore/issues/543
- https://github.com/semaphore-protocol/semaphore/issues/544
- https://github.com/semaphore-protocol/semaphore/issues/545


packages/proof/src/generate-proof.ts
```diff
@@ -0,0 +1,74 @@
...
+export default async function generateProof(
+    identity: Identity,
+    group: Group,
+    message: BytesLike | Hexable | number | bigint,
+    scope: BytesLike | Hexable | number | bigint,
+    treeDepth?: number,
+    snarkArtifacts?: SnarkArtifacts
+): Promise<SemaphoreProof> {
+    const leafIndex = group.indexOf(identity.commitment)
+    const merkeProof = group.generateMerkleProof(leafIndex)
```
vplasencia:
```suggestion
    const merkleProof = group.generateMerkleProof(leafIndex)
```

vplasencia:
Issues regarding the Merkle Tree circuit:
 1. When the parent of an element is itself.
    - Example: tree with members 1, 2, 3, 4, 5 and the index 4 which is the element 5.
    - This issue is related to this other issue: Generating the path indices for the merkle proof (https://github.com/semaphore-protocol/semaphore/blob/feat/semaphore-v4/packages/proof/src/generate-proof.ts#L46-L52).
        - The path indices and siblings are not correct for the circuit.
2. When the depth of the circuit is the same as the MAX_DEPTH.
    - To solve this we could add something like this code before `out <== root;` at the end.
        ``` circom
        var a = IsEqual()([depth, MAX_DEPTH]);

       roots[MAX_DEPTH] <== a * nodes[MAX_DEPTH];

       root += roots[MAX_DEPTH];
        ```
        And the roots signal should be `signal roots[MAX_DEPTH + 1];`

cedoor:
> Issues regarding the Merkle Tree circuit:

> 1. When the parent of an element is itself.

What happens exactly? Could you add a test for it, or share the code you used here?

> 2. When the depth of the circuit is the same as the MAX_DEPTH.

Good catch! I feel like this should be done here: https://github.com/privacy-scaling-explorations/zk-kit/blob/main/packages/circuits/circom/binary-merkle-root.circom. What do you think?

vplasencia:
@cedoor Yes, sure, I will create an example of test for it. Some changes should be in the zk-kit package and others in the semaphore proof package. For 2, yes, all the changes are in zk-kit.

vplasencia:
> 1. When the parent of an element is itself.

Actually, this issue is related to variable naming in the Merkle Tree circuit because the name depth works for the IMT but not for the LeanIMT.

A test case could be to use the leaves:

```
BigInt(1), BigInt(2), BigInt(3), BigInt(4), BigInt(5)
```

And generate the merkle proof of the position 4.

Replace this part (https://github.com/privacy-scaling-explorations/zk-kit/blob/main/packages/circuits/tests/binary-merkle-root.test.ts#L12-L20) with this:

```ts
    const leaf = BigInt(5)

    for (let i = 1; i < 6; i += 1) {
        tree.insert(BigInt(i))
    }

    const { siblings, index } = tree.generateProof(4)
```
With that, the test will not pass.

To pass the test, we can add `const siblingsLength = siblings.length` here:

```ts
    const indices: number[] = []
    const siblingsLength = siblings.length

    for (let i = 0; i < MAX_DEPTH; i += 1) {
        indices.push((index >> i) & 1)

        if (siblings[i] === undefined) {
            siblings[i] = BigInt(0)
        }
    }
```

And update the input:

```ts
    const INPUT = {
        leaf,
        depth: siblingsLength,
        indices,
        siblings
    }
```

To solve this issue, I think that it is better to change the name of the variable depth [here](https://github.com/privacy-scaling-explorations/zk-kit/blob/main/packages/circuits/circom/binary-merkle-root.circom#L17) and use another to refer to the siblings' length.

cedoor:
@vplasencia 

> Actually, this issue is related to variable naming in the Merkle Tree circuit because the name depth works for the IMT but not for the LeanIMT

The LeanIMT works a bit differently compared to the old IMT. The Merkle proof could contain a different index and number of siblings due to the tree properties. I'm not sure if the variable name should be something different than `depth`, as in most cases it _is_ the actual tree depth, and the circuit also supports the IMT in theory.

The `@semaphore-protocol/proof` package uses the sibling list length anyway, which is the actual depth in most cases:

https://github.com/semaphore-protocol/semaphore/blob/406bbc4f7d99fa8f1f56fccd5075b5c44127141f/packages/proof/src/generate-proof.ts#L31

It could be different when the leaf is on the right side of the tree.

Proposal:

https://github.com/privacy-scaling-explorations/zk-kit/blob/main/packages/circuits/circom/binary-merkle-root.circom could still use depth, as it's not necessary related to the LeanIMT.

Semaphore could use other names, like `treeProofLength`, `treeProofIndices` and `treeProofSiblings`.

vplasencia:
@cedoor 

Regarding this:
https://github.com/semaphore-protocol/semaphore/pull/480#issuecomment-1875110428

I like this proposal 👍.

packages/contracts/contracts/Semaphore.sol
```diff
@@ -150,15 +104,15 @@ contract Semaphore is ISemaphore, SemaphoreGroups {
...
-        uint256 signal,
-        uint256 nullifierHash,
-        uint256 externalNullifier,
+        uint256 nullifier,
+        uint256 message,
+        uint256 scope,
         uint256[8] calldata proof
-    ) external override {
+    ) external override onlyExistingGroup(groupId) {
         uint256 merkleTreeDepth = getMerkleTreeDepth(groupId);
```
numtel: This merkle tree depth is 0 when there's only 1 group member. The merkle root is simply the identity commitment at that time.

This just needs to be changed to `getMerkleTreeSize` instead and it will  work as intended.

See the [definition in the zk-kit imt.sol repo](https://github.com/privacy-scaling-explorations/zk-kit/blob/main/packages/imt.sol/contracts/internal/InternalLeanIMT.sol#L8)

.github/workflows/production.yml
```diff
@@ -63,7 +63,14 @@ jobs:
             - name: Install Node.js
               uses: actions/setup-node@v1
               with:
-                  node-version: 16.x
+                  node-version: "18"
```
0xjei: Should we switch to latest LTS version of Node?

v20 ( LTS )	8 months ago (18 Apr 2023) - 20.10.0 (22 Nov 2023)

packages/circuits/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 Ethereum Foundation
```
0xjei: nit: move to 2024?

packages/circuits/semaphore.circom
```diff
@@ -1,90 +1,25 @@
...
-    out <== poseidon.out;
-}
-
-template CalculateIdentityCommitment() {
-    signal input secret;
-
-    signal output out;
-
-    component poseidon = Poseidon(1);
+    merkleRoot <== BinaryMerkleRoot(MAX_DEPTH)(identityCommitment, merkleProofLength, merkleProofIndices, merkleProofSiblings);
```
0xjei, line 53: neat +1

packages/contracts/contracts/interfaces/ISemaphore.sol
```diff
@@ -3,26 +3,20 @@ pragma solidity 0.8.4;
 
 /// @title Semaphore contract interface.
 interface ISemaphore {
-    error Semaphore__CallerIsNotTheGroupAdmin();
+    error Semaphore__GroupHasNoMembers();
     error Semaphore__MerkleTreeDepthIsNotSupported();
     error Semaphore__MerkleTreeRootIsExpired();
     error Semaphore__MerkleTreeRootIsNotPartOfTheGroup();
     error Semaphore__YouAreUsingTheSameNillifierTwice();
```
0xjei:
```suggestion
    error Semaphore__YouAreUsingTheSameNullifierTwice();
```

packages/identity/README.md
```diff
@@ -49,8 +49,8 @@
     </h4>
 </div>
 
-| This library provides a class that can be used to create identities compatible with the Semaphore [circuits](https://github.com/semaphore-protocol/semaphore/tree/main/circuits). Each identity contains two secret values: _trapdoor_ and _nullifier_, and one public value: _commitment_. The Poseidon hash of the secret values is the identity secret, and its hash is the identity commitment. |
-| --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| This library provides a class that can be used to create identities compatible with the Semaphore [circuits](https://github.com/semaphore-protocol/semaphore/tree/main/circuits). Each identity contains an EdDSA private key, its public key, and the identity commitment, which is the Poseidon hash of the public key. |
```
0xjei: nit: maybe is worth to put an external link to EdDSA and Poseidon?

btw, definitely easier to read and understand!

packages/proof/README.md
```diff
@@ -86,18 +87,22 @@ import { utils } from "ethers"
...
 
 group.addMembers([...identityCommitments, identity.generateCommitment()])
 
-const fullProof = await generateProof(identity, group, externalNullifier, signal, {
-    zkeyFilePath: "./semaphore.zkey",
-    wasmFilePath: "./semaphore.wasm"
-})
+const fullProof1 = await generateProof(identity, group, message, scope)
+
+// You can also specify the maximum shaft depth supported by the proof.
```
0xjei: nit

```suggestion
// You can also specify the maximum tree depth supported by the proof.
```

packages/proof/src/get-snark-artifacts.browser.ts
```diff
@@ -0,0 +1,9 @@
+/* istanbul ignore file */
+import { SnarkArtifacts } from "./types"
+
+export default async function getSnarkArtifacts(treeDepth: number): Promise<SnarkArtifacts> {
+    return {
+        wasmFilePath: `https://semaphore.cedoor.dev/artifacts/${treeDepth}/semaphore.wasm`,
```
0xjei, line 15: nit: switch the host from your personal @cedoor to PSE one?

0xjei (changes requested):
Outstanding job :1st_place_medal: - Thank you @cedoor.

Left some comments and minor nits around.

packages/circuits/LICENSE
cedoor: Yes we should update all the licenses in V4

.github/workflows/production.yml
cedoor: I think it makes sense as v20 is LTS and then stable enough.

packages/contracts/contracts/interfaces/ISemaphore.sol
cedoor: Good catch, the other occurrences must also be updated.

packages/identity/README.md
cedoor: Yes, it makes sense 👍🏽

packages/proof/src/get-snark-artifacts.browser.ts
cedoor: This is temporary. We'll change host as soon as we run the ceremony and have the actual artifacts.

packages/contracts/contracts/Semaphore.sol
0xjei: Thank you @numtel. Please, see #535 to reflect your requested changes

.github/workflows/production.yml
0xjei: See [#536](https://github.com/semaphore-protocol/semaphore/pull/536)

packages/circuits/LICENSE
0xjei: See [#537](https://github.com/semaphore-protocol/semaphore/pull/537)

packages/contracts/contracts/interfaces/ISemaphore.sol
0xjei: See [#538](https://github.com/semaphore-protocol/semaphore/pull/538)

packages/identity/README.md
0xjei: See [#539](https://github.com/semaphore-protocol/semaphore/pull/539)

0xjei (approved):

packages/circuits/semaphore.circom
```diff
@@ -1,90 +1,25 @@
...
-include "./tree.circom";
+include "babyjub.circom";
+include "poseidon.circom";
+include "binary-merkle-root.circom";
 
-template CalculateSecret() {
-    signal input identityNullifier;
-    signal input identityTrapdoor;
+template Semaphore(MAX_DEPTH) {
+    signal input privateKey;
```
vplasencia: What about using the name `secret` or `secretScalar` instead of `privateKey` because it can be confusing with the private key of the identity, and they are not the same.

vplasencia (approved):
Great work! ✨

packages/circuits/semaphore.circom
cedoor: Yes, it makes sense!

cedoor: Please see: https://github.com/semaphore-protocol/semaphore/pull/480/commits/076e5de03c3646590e33f830dd75905e6fc0103d

PR #529, opened 2024-01-08
New Hardhat & Solidity versions

### LINKED ISSUE #524: Update Solidity pragma version and Hardhat in Semaphore V4

cedoor:
## Related Issue

Closes #524 

-   [ ] Yes
-   [x] No


packages/contracts/contracts/Semaphore.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity 0.8.4;
+pragma solidity 0.8.19;
```
vplasencia: Why not use `0.8.23` instead of `0.8.19`?

cedoor: It's the one Hardhat sets when you init a project. I think it's better to use it for compatibility. There are some warnings with the latest.

vplasencia: Which warnings? I tested with `0.8.23` and I didn't see any warnings.

cedoor: You're right :D I've probably seen them with Viem or some other configuration. Let me update it

vplasencia (approved):

PR #554, opened 2024-01-11
New theme colors and button style

### LINKED ISSUE #496: Set right colors in the `Button` style component of the Website

cedoor:
## Related Issue

Closes #496

-   [ ] Yes
-   [x] No

### DIFF
--- a/apps/website/src/styles/components/Button.ts
+++ b/apps/website/src/styles/components/Button.ts
@@ -1,6 +1,5 @@
 import { StyleFunctionProps, SystemStyleObject } from "@chakra-ui/theme-tools"
 import { font } from "../styles"
-import colors from "../colors"
 
 const Button = {
     baseStyle: {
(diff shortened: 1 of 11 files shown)


apps/website/src/styles/components/Button.ts
```diff
@@ -14,34 +13,53 @@ const Button = {
...
         solid: (props: StyleFunctionProps): SystemStyleObject => {
             const { colorScheme: c } = props
 
+            if (c === "primary") {
+                const bgGradient = "linear(to-r, primary.500, primary.800)"
+                const color = "white"
+
+                return {
+                    bgGradient,
+                    color,
```
vplasencia, line 28:
```suggestion
                    color,
                    transitionDuration:"200ms",
                    transitionTimingFunction: "linear",
```

What about adding something like this here?

vplasencia (approved):

apps/website/src/styles/components/Button.ts
vplasencia: Maybe do the same with all the other buttons?

cedoor: I like it, I've added them

vplasencia (approved):
Great work!

PR #639, opened 2024-02-18
chore(website): make caurosel horizontally scrollable (closes #562)

### LINKED ISSUE #562: Make caurosels horizontally scrollable in the website

intldds:
chore(website): make caurosel horizontally scrollable (closes #562)

 Added horizontal scrolling on the "Carousel.tsx" file

## Description

 Added horizontal scrolling on the "Carousel.tsx" file

 Feature (website)
 Carousel isn't scrolling horizontally (issue #562)
 Carousel should be scrolling horizontally now
 No

## Related Issue(s)

 #562 

## Other information

I've basically just added one line in the `Project Card` box:

`overflowX="auto"`

and it should work smoothly : )

## Checklist

-   [X] My code follows the style guidelines of this project
-   [X] I have performed a self-review of my code
-   [X] I have commented my code, particularly in hard-to-understand areas
-   [X] My changes generate no new warnings
-   [X] I have run `yarn prettier` and `yarn lint` without getting any errors


apps/website/src/components/Carousel.tsx
```diff
@@ -71,6 +71,7 @@ export default function Carousel({ title, sizes, type, ...props }: CarouselProps
             <HStack w="full" overflow="hidden">
                 <HStack
                     w="full"
+                    overflowX="auto"  // Enable horizontal scrolling
```
cedoor: It looks like the arrows buttons don't work properly after scrolling. I'm testing it with my laptop. Could you try it?

intldds: You are right -- I forgot to test the arrows buttons and they were not working.

I've added a new commit that should fix this :) (at least on my laptop in several browsers, both the arrows and the scrolling should be working now)

cedoor: Hey @intldds, they are working now. But when I scroll and then use the arrows the scrollbar changes its size and cannot be used to reach all the items anymore. Can you see it?

intldds: Hi @cedoor, the Right Arrow seems to be working fine with the desired behavior. However, the Left Arrow indeed is not working as supposed to. I would guess the problem is probably in the `previousProject` const, which is somehow being overridden by the scrolling functionality. 

I've tried making some changes but unfortunately couldn't make it work. It's a bit weird because the Right Arrow recognizes the index of the project and works fine, but when clicking on the Left Arrow it always assumes the index is 0 and takes me back to the last project. 

Both `previousProject` and `nextProject` are built in similar way, so I can't figure out why the scrollbar only causes this error on the left side.

vplasencia:
Hey @intldds! Thank you so much for all the work. What about using `onScroll={() => resetIndex()` here: https://github.com/semaphore-protocol/semaphore/pull/639/files#diff-ce27813f8a675d2fb0dab237071308614cd061c42bf95d38401bfe01ef405438R71 

We could have control every time the element is scrolled. Then we can set the value in a way that if it was scrolled and someone click on one of the arrows, the list will start from the beginning. 

Then, if someone wants to scroll, they can do it and if they want to click on the arrows they can do it too. 

What do you think?

intldds:
Hi @vplasencia! Sorry for taking so long with this, I had a surgery mid-March and took me a while to recover (feeling good now :))

I've followed your suggestion of adding the `onScroll={() => resetIndex()` property -- I think it should be working well now.

If you click on the left arrow, the list should start from the beginning. It feels intuitive, clean and easy to use -- let me know what you think and sorry once again for not replying sooner!

PR #652, opened 2024-02-25
Update `cli-template-monorepo-ethers`

vplasencia:
## Related Issue(s)

Closes #647

## Checklist

-   [x] My code follows the style guidelines of this project
-   [x] I have performed a self-review of my code
-   [x] I have commented my code, particularly in hard-to-understand areas
-   [x] I have made corresponding changes to the documentation
-   [x] My changes generate no new warnings
-   [x] I have run `yarn prettier` and `yarn lint` without getting any errors
-   [x] I have added tests that prove my fix is effective or that my feature works
-   [x] New and existing unit tests pass locally with my changes
-   [x] Any dependent changes have been merged and published in downstream modules

PR #653, opened 2024-02-26
Update `cli-template-monorepo-subgraph`

### LINKED ISSUE #648: Update `cli-template-monorepo-subgraph`

vplasencia:
## Related Issue(s)

Closes #648

## Checklist


0xjei (commented):
I'm getting the same errors as https://github.com/semaphore-protocol/semaphore/pull/652#pullrequestreview-1904065321

packages/cli-template-monorepo-subgraph/apps/contracts/.env.example
cedoor, line 1: We don't actually need this, we have the same env variables in the monorepo root folder. The `hardhat.config.ts` file could use `dotenv` like the boilerplate here: https://github.com/semaphore-protocol/boilerplate/blob/07b1365c4ad1b0320501f9254e7357e90e2bb6f7/apps/contracts/hardhat.config.ts#L14

packages/cli-template-monorepo-subgraph/apps/contracts/hardhat.config.ts
```diff
@@ -49,7 +42,7 @@ function getNetworks(): NetworksUserConfig {
     }
 }
 
-const hardhatConfig: HardhatUserConfig = {
+const config: HardhatUserConfig = {
     solidity: "0.8.23",
```
cedoor, line 49: It could also include this option: https://github.com/semaphore-protocol/boilerplate/blob/07b1365c4ad1b0320501f9254e7357e90e2bb6f7/apps/contracts/hardhat.config.ts#L55

packages/cli-template-monorepo-subgraph/apps/contracts/.gitignore
```diff
@@ -0,0 +1,71 @@
+node_modules
+.env
+
+# Hardhat files
+/cache
+/artifacts
+
+# TypeChain files
+/typechain
+/typechain-types
```
cedoor, line 7: These are the only folders to ignore here. The other ones are already included in the `.gitignore` file in the root folder.

packages/cli-template-monorepo-subgraph/apps/web-app/.gitignore
```diff
@@ -0,0 +1,36 @@
...
+
+# testing
+/coverage
+
+# next.js
+/.next/
+/out/
+
+# production
+/build
```
cedoor, line 8: We should only need these here.

packages/cli-template-monorepo-subgraph/apps/web-app/.env.development.local.example
cedoor, line 1: This is not an example. It should actually be the env file that NextJS uses for its dev env. `example` is usually used for ignored `.env` files that need to be copied before starting a project. Same comment for the other local env file.

packages/cli-template-monorepo-subgraph/apps/contracts/tsconfig.json
```diff
@@ -1,15 +1,11 @@
...
-        "module": "CommonJS",
-        "strict": true,
+        "target": "es2020",
+        "module": "commonjs",
         "esModuleInterop": true,
-        "outDir": "dist",
-        "typeRoots": ["node_modules/@types", "types"]
-    },
-    "include": ["scripts/**/*", "tasks/**/*", "test/**/*", "build/typechain/**/*", "types/**/*"],
-    "files": ["./hardhat.config.ts"]
```
cedoor: Not sure if we don't need these.

packages/cli-template-monorepo-subgraph/apps/contracts/test/Feedback.ts
```diff
@@ -40,28 +47,38 @@ describe("Feedback", () => {
...
+
+            for (const user of users) {
+                await feedbackContract.joinGroup(user.commitment)
+                group.addMember(user.commitment)
+            }
+
+            const feedback = encodeBytes32String("Hello World")
 
-            const proof = await generateProof(users[1], group, feedback, groupId)
+            const fullProof = await generateProof(users[1], group, feedback, groupId)
```
cedoor: After renaming `proof` to `points`, we started using `proof` rather than `fullProof`.

packages/cli-template-monorepo-subgraph/apps/contracts/package.json
```diff
@@ -1,42 +1,54 @@
...
+        "solhint": "^4.1.1",
+        "solidity-coverage": "^0.8.1",
+        "ts-node": "^10.9.2",
+        "typechain": "^8.3.0",
+        "typescript": "^5.3.3"
     },
     "dependencies": {
-        "@semaphore-protocol/contracts": "4.0.0-alpha.8"
-    }
+        "@semaphore-protocol/contracts": "^4.0.0-alpha.8"
```
cedoor: It would be better to not use `^` and specify the version here. `^` may introduce unexpected issues as it could use different versions.

packages/cli-template-monorepo-subgraph/apps/contracts/.prettierignore
cedoor, line 1: We don't really need this file in this case. There's already the `.prettierignore` file in the root folder.

packages/cli-template-monorepo-subgraph/apps/contracts/.eslintignore
```diff
@@ -8,17 +8,21 @@ yarn.lock
...
+artifacts
+cache
+typechain-types
 
 # types
 types
 
+# circuits
+circuits
+
```
cedoor: We shouldn't need to ignore these folders as they shouldn't exist.

packages/cli-template-monorepo-subgraph/apps/contracts/.eslintignore
```diff
@@ -8,17 +8,21 @@ yarn.lock
...
 # production
 dist
 build
-cache
-contract-artifacts
+docs
```
cedoor: This, `cli-template-*` and `snark-artifacts` don't exist either. No need to have theme here.

packages/cli-template-monorepo-subgraph/apps/contracts/.yarnrc.yml
cedoor, line 1: This monorepo already has a `.yarnrc.yml` file in the root folder.

packages/cli-template-monorepo-subgraph/apps/web-app/package.json
```diff
@@ -1,25 +1,29 @@
...
-        "@next/font": "13.0.3",
-        "@semaphore-protocol/core": "4.0.0-alpha.8",
-        "@semaphore-protocol/data": "4.0.0-alpha.8",
-        "@types/react": "18.0.25",
-        "@types/react-dom": "18.0.8",
-        "dotenv": "^16.0.3",
-        "ethers": "^6.4.0",
-        "next": "13.0.3",
+        "@semaphore-protocol/core": "^4.0.0-alpha.8",
+        "@semaphore-protocol/data": "^4.0.0-alpha.8",
```
cedoor: It would be better to remove `^` from these deps. I already added similar comments with more info about this.

packages/cli-template-monorepo-subgraph/package.json
```diff
@@ -16,28 +16,19 @@
         "README.md"
     ],
```
cedoor, line 13: I think some of these files can be removed now: `tsconfig.json`, `.eslintrc.json`, `.eslintignore`, `scripts`.

packages/cli-template-monorepo-subgraph/README.md
```diff
@@ -1,16 +1,30 @@
...
+cp ./apps/web-app/.env.development.local.example ./apps/web-app/.env.development.local
+```
+
+-   Copy the `.env.production.local.example` file as `.env.production.local`:
+
+```bash
+cp ./apps/web-app/.env.production.local.example ./apps/web-app/.env.production.local
+```
+
+Add your environment variables or run the app in a local network.
```
cedoor: They don't need to copy those files. Those variables are not secret. We can just explain they can update those env variables when the addresses or something else changes. I added other comments for those env files.

cedoor (changes requested):
Thanks 🙏🏽 

I think the comments I left here are the same as the ones I'd leave in the other PR: https://github.com/semaphore-protocol/semaphore/pull/652. Please consider this review as valid for that PR too.

packages/cli-template-monorepo-subgraph/apps/web-app/package.json
vplasencia: I get it for the contracts, but what would be the issue with the JS libraries if the JS libraries are following the semantic versioning?

cedoor: Just to have more control of the actual Semaphore code the app is using. With `^` it could use `4.1.0` or `4.12.0`. Minor versions could have important changes on JS libraries.

cedoor (approved):

PR #660, opened 2024-02-26
Adds `getGroupMembers` & `getGroupValidatedProofs` functions to `SemaphoreSubgraph`

### LINKED ISSUE #619: Add `getGroupMembers` & `getGroupValidatedProofs` functions to `SemaphoreSubgraph`

zkfriendly:
## Related Issue(s)
Closes #619 

## Checklist

-   [x] I have run `yarn prettier` and `yarn lint` without getting any **new** errors

### DIFF
--- a/packages/data/src/subgraph.ts
+++ b/packages/data/src/subgraph.ts
@@ -209,7 +209,7 @@ export default class SemaphoreSubgraph {
      */
     async getGroupMembers(groupId: string): Promise<string[]> {
         const group = await this.getGroup(groupId, { members: true }) // parameters are checked inside getGroup
-        return group.members ?? []
+        return group.members!
     }
 
     /**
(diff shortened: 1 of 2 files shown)


packages/data/src/subgraph.ts
```diff
@@ -202,6 +202,26 @@ export default class SemaphoreSubgraph {
...
     }
 
+    /**
+     * Returns a list of group members.
+     * @param groupId Group id.
+     * @returns Group members.
+     */
+    async getGroupMembers(groupId: string): Promise<string[]> {
+        const group = await this.getGroup(groupId, { members: true }) // parameters are checked inside getGroup
+        return group.members ?? []
```
cedoor: `group.members` should be either an empty list or a list of members. So `?? []` may not be necessary. Have you checked it?

cedoor (commented):
Thanks for this PR @zkfriendly !

packages/data/src/subgraph.ts
zkfriendly: right `?? []` is not needed. It could be: `return group.members!`. And I suppose the same is true for `getGroupValidatedProofs` ?

cedoor: Yes, it should work in the same way 👍🏽

cedoor (approved):

Issue #483, opened 2023-12-18
Add more detailed documentation to Semaphore V4 contracts and libraries


cedoor:
**Tasks**:

- [x] Circuits 
- [x] `@semaphore-protocol/contracts`
- [x] Core packages
    - [x] `@semaphore-protocol/identity`
    - [x] `@semaphore-protocol/group`
    - [x] `@semaphore-protocol/proof`
- [x] `@semaphore-protocol/utils`
- [x] `@semphore-protocol/data` https://github.com/semaphore-protocol/semaphore/pull/768
- [x] `@semaphore-protocol/hardhat` https://github.com/semaphore-protocol/semaphore/pull/772
- [x] `@semaphore-protocol/cli` https://github.com/semaphore-protocol/semaphore/pull/771

PR #663, opened 2024-02-28
More documentation on Identity lib code

cedoor:
## Description

This PR adds more documentation to the `Identity` class and updates the TypeDoc link. It also removes the class method to verify signatures as the static method can be used instead.

## Related Issue(s)

#483 

## Checklist

### DIFF
--- a/packages/identity/src/identity.ts
+++ b/packages/identity/src/identity.ts
@@ -53,7 +53,7 @@ export default class Identity {
 
     /**
      * Returns the private key.
-     * @returns The private key as a bignumber-ish.
+     * @returns The private key as a {@link https://zkkit.pse.dev/types/_zk_kit_utils.BigNumberish.html|BigNumberish}.
      */
     public get privateKey(): BigNumberish {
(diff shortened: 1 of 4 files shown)


packages/identity/src/identity.ts
```diff
@@ -10,65 +10,111 @@ import {
...
-
         this._publicKey = derivePublicKey(privateKey)
-
         this._commitment = poseidon2(this._publicKey).toString()
     }
 
     /**
      * Returns the private key.
-     * @returns The private key.
+     * @returns The private key as a bignumber-ish.
```
vplasencia: We could add a link to the bignumber-ish type here and the same for the Baby Jubjub point.

vplasencia (changes requested):

packages/identity/src/identity.ts
cedoor: Yes, I've added those links.

vplasencia (approved):

PR #664, opened 2024-02-28
More documentation on proof lib code

cedoor:
## Description

This PR adds more documentation to the proof package functions and updates the TypeDoc link.

## Related Issue(s)

#483

## Checklist

### DIFF
--- a/packages/proof/README.md
+++ b/packages/proof/README.md
@@ -18,7 +18,7 @@
     <a href="https://npmjs.org/package/@semaphore-protocol/proof">
         <img alt="Downloads" src="https://img.shields.io/npm/dm/@semaphore-protocol/proof.svg?style=flat-square" />
     </a>
-    <a href="https://js.semaphore.pse.dev/proof">
+    <a href="https://js.semaphore.pse.dev/modules/_semaphore_protocol_proof">
         <img alt="Documentation typedoc" src="https://img.shields.io/badge/docs-typedoc-744C7C?style=flat-square">
     </a>
     <a href="https://eslint.org/">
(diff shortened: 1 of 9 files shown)


packages/proof/src/generate-proof.ts
```diff
@@ -9,7 +9,17 @@ import toBigInt from "./to-bigint"
 import { BigNumberish, SemaphoreProof, SnarkArtifacts } from "./types"
 
 /**
- * Generates a Semaphore proof.
+ * It generates a Semaphore proof, i.e. a zero-knowledge proof that an identity that
+ * is part of a group has shared an anonymous message.
+ * The message may be any arbitrary user-defined value (e.g. a vote), or the hash of that value.
+ * The scope is a value used like a topic on which users can generate a valid proof only once,
+ * for example the id of an election in which voters can only vote once.
+ * The hash of the identity's scope and private key is called a nullifier and can be
```
vplasencia:
```suggestion
 * The hash of the identity's scope and secret scalar is called a nullifier and can be
```

packages/proof/src/index.ts
```diff
@@ -1,7 +1,6 @@
 import { packGroth16Proof, unpackGroth16Proof } from "@zk-kit/utils"
 import generateProof from "./generate-proof"
-import getSnarkArtifacts from "./get-snark-artifacts.node"
 import verifyProof from "./verify-proof"
 
 export * from "./types"
-export { generateProof, getSnarkArtifacts, packGroth16Proof, unpackGroth16Proof, verifyProof }
```
vplasencia, line 7: It would be nice to have `getSnarkArtifacts` for node as a utility function to make it easier for devs to download the artifacts in case they need it because the automatic download does not work for them.

vplasencia (changes requested):
Looks great ✨

Added some comments.

packages/proof/src/index.ts
cedoor: If the automatic download doesn't work, that function won't work used externally either I think :/

I think it can actually be useful if you can define your own URLs.

vplasencia: Yes, true. It should be removed from here. Maybe adding this feature somewhere else to do it.

vplasencia (approved):

packages/proof/README.md
```diff
@@ -49,8 +49,8 @@
...
-| ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| This library provides utility functions to generate and verify Semaphore proofs compatible with the Semaphore [circuits](https://github.com/semaphore-protocol/semaphore/tree/main/circuits). Generating valid zero-knowledge proofs requires files that can only be obtained in an attested trusted-setup ceremony. |
```
0xjei, line 52: "... attested trusted-setup ceremony."

What do you mean?

0xjei (commented):
Lovely work. Reading some new comments on top of the functions is literally a pleasure.

packages/proof/src/index.ts
cedoor: That can be part of another issue. It could be worth checking the code again and seeing if some other functions can be moved to the utils package.

packages/proof/README.md
cedoor: I don't remember if I wrote it but it should mean "an official trusted-setup". Sounds like a good term to me.

PR #674, opened 2024-02-29
Download the correct CLI Template files

### LINKED ISSUE #673: Download the correct CLI template files
### Describe the bug

- The `cli-template-contracts-hardhat` is not downloading the `.yarn` folder and the `yarnrc.yml` file.

- The `cli-template-monorepo-ethers` and `cli-template-monorepo-hardhat` templates are downloading folders that are not necessary like: `typechain-types`, `artifacts` and `cache`.

vplasencia:
## Related Issue(s)

Closes #673

## Checklist

### DIFF
--- a/packages/cli-template-contracts-hardhat/package.json
+++ b/packages/cli-template-contracts-hardhat/package.json
@@ -32,7 +32,7 @@
         "test:report-gas": "REPORT_GAS=true hardhat test",
         "test:coverage": "hardhat coverage",
         "typechain": "hardhat typechain",
-        "prepublish": "tar --exclude typechain-types --exclude artifacts --exclude cache  --exclude node_modules -czf files.tgz .gitignore .yarn .yarnrc.yml",
+        "prepublish": "tar -czf files.tgz .gitignore .yarn .yarnrc.yml",
         "lint": "eslint . --ext .js,.ts && solhint 'contracts/**/*.sol'",
         "prettier": "prettier -c --plugin=prettier-plugin-solidity .",
         "prettier:write": "prettier -w --plugin=prettier-plugin-solidity ."
--- a/packages/cli-template-contracts-hardhat/package.json
+++ b/packages/cli-template-contracts-hardhat/package.json
@@ -32,7 +32,7 @@
         "test:report-gas": "REPORT_GAS=true hardhat test",
         "test:coverage": "hardhat coverage",
         "typechain": "hardhat typechain",
-        "prepublish": "tar -czf files.tgz .gitignore",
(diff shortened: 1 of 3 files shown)


packages/cli-template-contracts-hardhat/package.json
```diff
@@ -32,7 +32,7 @@
         "test:report-gas": "REPORT_GAS=true hardhat test",
         "test:coverage": "hardhat coverage",
         "typechain": "hardhat typechain",
-        "prepublish": "tar -czf files.tgz .gitignore",
+        "prepublish": "tar --exclude typechain-types --exclude artifacts --exclude cache  --exclude node_modules -czf files.tgz .gitignore .yarn .yarnrc.yml",
```
cedoor: I'm not sure if we need to explicitly exclude those folders. I've just tried to run `tar -czf files.tgz .gitignore` and even if there are those folders (typechain-types, artifacts, node_modules), they've not been included.

vplasencia: If you run `semaphore create` and select the `monorepo-ethers` template for example, you will see the `artifacts` and `typechain-types` there inside `apps/contracts`. Then we can include those files to make sure they will never be there or make sure that the files are not there when releasing a new version of the packages. 

I think that with automatic releases, we wouldn't need to exclude the files.

cedoor: Sounds right for the monorepos, but we shouldn't need to exclude them in the contracts-hardhat template right?

vplasencia: Yes, you're right. That's not necessary for this specific template, just for the others. I will update it.

packages/cli-template-contracts-hardhat/package.json
```diff
@@ -32,7 +32,7 @@
         "test:report-gas": "REPORT_GAS=true hardhat test",
```
cedoor, line 30: I think `.yarn` and `.yarnrc.yml` can also be removed from the files list.

cedoor (changes requested):

packages/cli-template-contracts-hardhat/package.json
vplasencia: The project is using `yarn` and if these files are not downloaded, devs would need to at least create a `.yarnrc.yml` file. We could remove them when we adapt the templates to use other package manager by default like `npm` for example. What do you think?

For this specific template we could remove the `.yarn` folder and modify the `yarnrc.yml` file to have only `nodeLinker: node-modules`.

cedoor: Sorry, I meant from the package.json's `files` field list.

cedoor (approved):

PR #703, opened 2024-03-14
Use an incremental group id when creating a group in contracts

### LINKED ISSUE #682: createGroup can be frontrun
A malicious actor can monitor calls of createGroup(groupId, admin) and frontrun it with createGroup(groupId, admin1). This prevents new group's creation. If we adopt an incremental groupId, once groupId increases by one each time createGroup is called, we can prevent this from happening

https://github.com/semaphore-protocol/semaphore/blob/8eb19e83fda62644872b2fcfbd85011d3b2c21e2/packages/contracts/contracts/Semaphore.sol#L27
https://github.com/semaphore-protocol/semaphore/blob/8eb19e83fda62644872b2fcfbd85011d3b2c21e2/packages/contracts/contracts/Semaphore.sol#L34

vplasencia:
## Related Issue(s)

Closes #682 

## Checklist


packages/contracts/contracts/Semaphore.sol
```diff
@@ -17,21 +17,27 @@ contract Semaphore is ISemaphore, SemaphoreGroups {
...
     mapping(uint256 => Group) public groups;
 
+    /// @dev Counter to assign the group id.
+    uint256 public groupCounter;
+
     /// @dev Initializes the Semaphore verifier used to verify the user's ZK proofs.
     /// @param _verifier: Semaphore verifier addresse.
     constructor(ISemaphoreVerifier _verifier) {
         verifier = _verifier;
+        groupCounter = 0;
```
cedoor: This shouldn't be necessary. `groupCounter` is 0 by default.

packages/contracts/contracts/Semaphore.sol
```diff
@@ -17,21 +17,27 @@ contract Semaphore is ISemaphore, SemaphoreGroups {
     /// @dev Gets a group id and returns the group parameters.
     mapping(uint256 => Group) public groups;
 
+    /// @dev Counter to assign the group id.
```
cedoor:
```suggestion
    /// @dev Counter to assign an incremental id to the groups.
```

packages/contracts/.prettierignore
cedoor, line 1: Why do we need this separate `.prettierignore` file?

cedoor (changes requested):

packages/contracts/.prettierignore
vplasencia: Thanks! I updated the prettier config in contracts.

cedoor (approved):
LGTM great!!

Side note, we could add a new `createGroup` function without any parameter, where the admin is msg.sender.

vplasencia:
Hey @cedoor I think that's a great idea, we could create a new issue. This will make it easier to use the function. There are a lot of cases where the admin is `msg.sender`.

0xDatapunk (approved):

0xbok (approved):

PR #719, opened 2024-03-21
Add deployed contracts to utils package

### LINKED ISSUE #715: Add the deployed contracts to the utils package
### Description

Add the deployed contracts to the Semaphore utils package so that other packages and projects can access this info.

The idea is to create a json file with the network's info (chain id, etc.) and the deployed contract addresses. The `supported-networks.ts` would import that json file and export it as a js object.

Deployed contracts json file: https://github.com/semaphore-protocol/semaphore/blob/main/packages/contracts/deployed-contracts.json

cedoor:
## Description

This PR centralizes the management of the networks and the deployed contracts supported by Semaphore, making it easier to deploy and support new networks. Here the steps:
1. Go to `packages/utils` and add a new network to the `src/networks.ts` file
2. Build the package with `yarn build`
3. Go to `packages/contracts` and deploy the Semaphore contracts with `yarn deploy --network <new-network>`
4. Run `yarn prettier:write` and commit your changes

Additionally, it removes OpenZeppelin Defender, as we cannot use it to deploy contracts due to the free subscription limits. It will be necessary to open another PR to integrate a mechanism to deploy contracts deterministically without Open Zeppelin.

## Related Issue(s)

Closes #715

## Checklist


packages/utils/src/networks.ts
```diff
@@ -0,0 +1,71 @@
...
+export function isSupportedNetwork(supportedNetwork: string): boolean {
+    return Object.keys(supportedNetworks).includes(supportedNetwork)
+}
+
+export function getHardhatNetworks(account?: string) {
+    if (!account) {
+        return {}
+    }
+
+    const supportedNetworks2 = JSON.parse(JSON.stringify(supportedNetworks))
```
vplasencia: What about using a different name for `supportedNetworks2 `. It could be `supportedNetworksObj` or `supportedNetworksObject`.

packages/utils/src/networks.ts
```diff
@@ -0,0 +1,71 @@
+import deployedContracts from "./deployed-contracts.json"
+
+// List of Semaphore supported networks.
+export const supportedNetworks = {
```
vplasencia: What if we have a folder called `networks` with everything related to the networks module there? 

Also, what about creating a file for `supportedNetworks ` called `supported-networks.ts`. We could import `supportedNetworks` in this `networks.ts` file. 

Then, when we add a new network, we just modify these two files with just network info without code: `deployed-contracts.json` and `supported-networks.ts`.

The `networks` folder would have: `deployed-contracts.json`, `supported-networks.ts`, `networks.ts`, etc. 

The idea would be to have everything related to a module inside a file or if there are several files, inside a folder since the modules are not related to each other.

packages/utils/tests/index.test.ts
```diff
@@ -1,13 +1,25 @@
 import { encodeBytes32String } from "ethers/abi"
 import { toBigInt } from "ethers/utils"
-import { supportedNetworks } from "../src"
 import decodeMessage from "../src/decode-message"
+import { getDeployedContract, isSupportedNetwork } from "../src/networks"
 
 describe("Utils", () => {
```
vplasencia, line 6: What about creating a `describe` per module? Then the describe sections we already have would be inside their own module. We could also have a test file per module.

vplasencia (commented):
Hey! Great work! Thank you so much.

Just left some comments.

packages/utils/src/networks.ts
cedoor: I could use `supportedNetworksCopy` if that sounds better

packages/utils/src/networks.ts
cedoor: I updated it. Let me know if it makes sense. Also, added documentation and tests.

packages/utils/tests/index.test.ts
cedoor: Makes sense. I created a module for `networks`, let's do the rest in another PR.

vplasencia (approved):

PR #751, opened 2024-04-24
ci: fix circom deps installation steps

sripwoud:
The circom deps in the pull requests workflow can't be install because there is an issue with their ubuntu repository (probably a temporary issue on the microsoft side related to their GPG keys).

As a temporary fix we can simply delete the microsoft source file in the workflow environment.
Besides, as long as the integrity checks of the packages we get from the microsoft repo are broken, I think it is fine to not fetch packages from their repo at all.
The drawback of not getting their softwares updates does not matter so much for us, as this is a ci environment we just use for running tests.

https://github.com/orgs/community/discussions/120966
https://stackoverflow.com/a/78377916

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -82,8 +82,11 @@ jobs:
                   cache: yarn
 
             # https://github.com/iden3/circuits/blob/8fffb6609ecad0b7bcda19bb908bdb544bdb3cf7/.github/workflows/main.yml#L18-L22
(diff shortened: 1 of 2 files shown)


sripwoud:
~~or we can just wait for microsoft to fix it~~
probably not, this seems to be a recurring issue at microsoft.
https://stackoverflow.com/a/78377916

So we could consider always skipping their repo in the ci?

cedoor (approved):

PR #768, opened 2024-05-06
Improve documentation for `@semaphore/data` package

0xjei:
## Description
This PR aims at improving the documentation (code / README) for the `@semaphore/data` package.

## Related Issue(s)
see #483 

## Checklist

### DIFF
--- a/packages/data/README.md
+++ b/packages/data/README.md
@@ -2,7 +2,7 @@
     <h1 align="center">
         Semaphore data
     </h1>
-    <p align="center">A library to query Semaphore contracts.</p>
+    <p align="center">A library for querying Semaphore smart contract.</p>
 </p>
 
 <p align="center">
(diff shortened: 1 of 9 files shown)


packages/data/README.md
```diff
@@ -2,7 +2,7 @@
     <h1 align="center">
         Semaphore data
     </h1>
-    <p align="center">A library to query Semaphore contracts.</p>
+    <p align="center">A library for querying Semaphore smart contract.</p>
```
cedoor, line 5: Could you update the `package.json` description to reflect this change?

packages/data/README.md
```diff
@@ -70,130 +70,121 @@ yarn add @semaphore-protocol/data
...
-const supportedNetworks = getSupportedNetworks()
-```
-
-\# **new SemaphoreSubgraph**(networkOrSubgraphURL: SupportedNetwork | ValueOf\<SupportedNetwork> | string = "sepolia"): _SemaphoreSubgraph_
+**Initialize a Semaphore Subgraph instance**
 
 ```typescript
 import { SemaphoreSubgraph } from "@semaphore-protocol/data"
 
+const supportedNetworks = getSupportedNetworks()
```
cedoor: We need to update this I think. That function has been moved to the `@semphore-protocol/utils` package. It's an object now. You can check that package to know more. There's no enum or something like `supportedNetworks.arbitrum` tho, so I'm not sure if we should use it here. A `"arbitrum"` string should be the right way, unless we don't add an enum to the utils package.

packages/data/src/checkParameter.ts
```diff
@@ -1,8 +1,11 @@
...
- * @param type Expected parameter type.
+ * Validates the type of a given parameter against an expected type.
+ * Throws a {@link TypeError} if the validation fails.
+ * This function is useful for ensuring that function arguments conform to expected types at runtime.
+ * @param value The value of the parameter to check.
+ * @param name The name of the parameter, used in the error message for easier debugging.
+ * @param type The expected JavaScript type as a string (e.g., 'string', 'number', 'object').
+ * @throws {TypeError} Throws an error if the type of `value` does not match the `type`.
  */
 export default function checkParameter(value: any, name: string, type: string) {
```
cedoor, line 10: Not sure if this should be part of this PR, but we have a ZK-Kit utils package to check parameters now. We should use it instead of this function.

cedoor (changes requested):

packages/data/src/checkParameter.ts
0xjei: hmmm I don't think so - we can open a small issue and amend it later.

cedoor (approved):

PR #790, opened 2024-05-20
docs: add link to API Reference (typedoc)

sripwoud:
- fix rendering of image in typedoc website's homepage
  Currently we only see the `img` `alt` description
  image
- shorten `docs` script (use config options in `typedoc.js` instead)
- optimize docs workflow: restrict the changed files that trigger it
- add link to `js.semaphore.pse.dev` in main README

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,13 +2,19 @@ name: docs
 
 on:
     push:
-        branches:
-            - main
+        branches: [main]
+        paths:
+            [
+                "apps/docs/**",
+                "packages/data/**",
+                "packages/group/**",
+                "packages/identity/**",
+                "packages/proof/**",
+                "packages/utils/**"
+            ]
 
 jobs:
     gh-pages:
-        runs-on: ubuntu-latest
-
         steps:
             - uses: actions/checkout@v4
 
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 <p align="center">
     <h1 align="center">
       <picture>
-        <source media="(prefers-color-scheme: dark)" srcset="https://github.com/semaphore-protocol/.github/blob/main/assets/semaphore-logo-light.svg">
-        <source media="(prefers-color-scheme: light)" srcset="https://github.com/semaphore-protocol/.github/blob/main/assets/semaphore-logo-dark.svg">
-        <img width="250" alt="Semaphore icon" src="https://github.com/semaphore-protocol/.github/blob/main/assets/semaphore-logo-dark.svg">
+        <source media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/semaphore-protocol/.github/main/assets/semaphore-logo-light.svg">
+        <source media="(prefers-color-scheme: light)" srcset="https://raw.githubusercontent.com/semaphore-protocol/.github/main/assets/semaphore-logo-dark.svg">
+        <img width="250" alt="Semaphore icon" src="https://raw.githubusercontent.com/semaphore-protocol/.github/main/assets/semaphore-logo-dark.svg">
       </picture>
     </h1>
 </p>
@@ -55,6 +55,10 @@
         <span>&nbsp;&nbsp;|&nbsp;&nbsp;</span>
         <a href="https://semaphore.pse.dev/discord">
             🗣️ Chat &amp; Support
+        </a>
+        <span>&nbsp;&nbsp;|&nbsp;&nbsp;</span>
+        <a href="https://js.semaphore.pse.dev">
+            💻 API Reference
         </a>
     </h4>
 </div>
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
         "lint": "eslint . --ext .js,.ts,.tsx && yarn workspace semaphore-contracts lint",
         "format": "prettier -c . && yarn workspace semaphore-docs format",
         "format:write": "prettier -w . && yarn workspace semaphore-docs format:write",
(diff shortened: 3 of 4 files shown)


.github/workflows/docs.yml
```diff
@@ -2,13 +2,19 @@ name: docs
 
 on:
     push:
-        branches:
-            - main
+        branches: [main]
```
sripwoud, line 5: let's trigger that workflow only when relevant files are changed

README.md
```diff
@@ -1,9 +1,9 @@
 <p align="center">
     <h1 align="center">
       <picture>
-        <source media="(prefers-color-scheme: dark)" srcset="https://github.com/semaphore-protocol/.github/blob/main/assets/semaphore-logo-light.svg">
-        <source media="(prefers-color-scheme: light)" srcset="https://github.com/semaphore-protocol/.github/blob/main/assets/semaphore-logo-dark.svg">
-        <img width="250" alt="Semaphore icon" src="https://github.com/semaphore-protocol/.github/blob/main/assets/semaphore-logo-dark.svg">
+        <source media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/semaphore-protocol/.github/main/assets/semaphore-logo-light.svg">
```
sripwoud, line 4: rendering of images work fine when using non raw urls in the main readme on github. But not when building it as html with typedoc. In that case it is necessary to use raw.githubusercontent.

See deployed docs: we don't see the logo on homepage, but the `alt` description instead

image

README.md
```diff
@@ -56,6 +56,10 @@
         <a href="https://semaphore.pse.dev/discord">
             🗣️ Chat &amp; Support
         </a>
+        <span>&nbsp;&nbsp;|&nbsp;&nbsp;</span>
```
sripwoud, line 59: There was no link to `js.semaphore.pse.dev` anywhere. Was it intentional?
If yes and the intention is to not use the typedocs anymore, we could delete the corresponding GH workflows, npm script and config file.

package.json
```diff
@@ -20,7 +20,7 @@
         "lint": "eslint . --ext .js,.ts,.tsx && yarn workspace semaphore-contracts lint",
         "format": "prettier -c . && yarn workspace semaphore-docs format",
         "format:write": "prettier -w . && yarn workspace semaphore-docs format:write",
-        "docs": "typedoc --cname js.semaphore.pse.dev --githubPages true",
+        "docs": "typedoc",
```
sripwoud, line 23: `cname` can be provided via config file.
And `githubPages` default to true
see
https://typedoc.org/schema.json

.github/workflows/docs.yml
```diff
@@ -2,13 +2,19 @@ name: docs
...
+        paths:
+            [
+                "apps/docs/**",
+                "packages/data/**",
+                "packages/group/**",
+                "packages/identity/**",
+                "packages/proof/**",
+                "packages/utils/**"
+            ]
```
sripwoud, line 14: let's trigger that workflow only when relevant files are changed

typedoc.js
```diff
@@ -2,15 +2,16 @@ const fs = require("fs")
...
     "cli-template-monorepo-ethers",
     "cli-template-monorepo-subgraph",
-    "core",
-    "circuits",
     "contracts",
+    "core",
     "hardhat",
-    "cli"
+    "heyauthn"
 ]
```
sripwoud: sorting and adding `heyauthn` to exclude list

gatleas17 (approved):

gatleas17 (approved):

README.md
0xjei: very neat - tysm for the explanation!

README.md
0xjei: i remember that typedocs is one of the things we'd like to keep for Semaphore at 99% but circle back to @cedoor or @vplasencia for more details!

package.json
0xjei: amazing, +1

0xjei (approved):
opti & neat - lgtm!

typedoc.js
vplasencia: Hey @sripwoud 

I think we can remove `heyauthn` since it's not part of this repository anymore. It's now part of the Semaphore extensions repository: https://github.com/semaphore-protocol/extensions

README.md
vplasencia: Hey @sripwoud 

Typedoc is very important for the project. There is a typedoc link in every package with type documentation. I agree that it would be better to have it in the main Readme file as well. 

Should it be part of the end of the Packages section instead of the beginning of the file? What do you all think?

vplasencia (changes requested):
Hey @sripwoud 

Thank you very much for this PR 🙏

Just left some comments.

typedoc.js
sripwoud: https://github.com/semaphore-protocol/semaphore/pull/790/commits/6d87016423a1ca7f316944141f24f992ae71563d

vplasencia (approved):

README.md
sripwoud: What about both? in the main README and in the packages' READMEs?
In any case I personally prefer to see such a doc link at the top of a README: the more I have to scroll down, the more likely I'll miss that link :wink: (I had missed the typedoc links [here](https://github.com/semaphore-protocol/semaphore/blob/main/packages/group/README.md#-usage)).
I would also add it somewhere on https://docs.semaphore.pse.dev/

vplasencia: Makes sense. Thanks. Feel free to create an issue to add it to https://docs.semaphore.pse.dev/

PR #803, opened 2024-06-13
 fix(identity)!: update type of private key 

### LINKED ISSUE #799: Generating a new identity with a user-provided private key results in inconsistent behavior
**Describe the bug**
Initialize two fresh Semaphore identities: a `new Identity()` with empty args, and another `new Identity(<private key>)` passing in the private key generated for the first identity. These two identities will have different public keys and commitments, though they should be identical.

**To Reproduce**

1. Initialize a fresh Semaphore identity: `const id1 = new Identity()`
2. Copy the private key of this identity: `const privkey = id1.privateKey`.
3. Initialize a second Semaphore identity, passing in the private key as an argument to constructor: `const id2 = new Identity(privkey)`. 

**Expected behavior**
```
console.log(id1.privateKey === id2.privateKey) // expect true, got true
console.log(id1.commitment === id2.commitment) // expect true, got false
```

**Screenshots**
`id1` in this screenshot was generated as a fresh Semaphore identity with no args passed into constructor. `id2` is generated with `id1`'s private key passed as argument to constructor. Both of these Identities appear to have the same privateKey, but have different publicKey and commitment values.

image

**Technologies (please complete the following information):**

-   Node.js version: v22.1.0
-   NPM version: 10.7.0
-   Solidity version: n/a

**Additional context**

Prepending the private key with `0x` prefix does not solve the issue: for example, run `const id3 = new Identity('0x' + privkey)`. Now, `id3`'s commitment and publicKey will be different from both the freshly generated Semaphore ID as well as the identity generated by passing in the first privateKey as a constructor argument.

cedoor:
## Description

This fixes #799 by updating the `@zk-kit/eddsa-poseidon` dependency and adds two methods to export/import encoded (base64) private keys.

The type of the private key in the Identity constructor has also been changed. Now, it must be one of the following: Uint8Array, Buffer, string, i.e. bytes or text, exactly the same as the types supported by `@zk-kit/eddsa-poseidon`.

This should eliminate the ambiguity of having a string that can be either text or hexadecimal, as it was before. 

## Related Issue(s)

Fixes #799

## Checklist

### DIFF
--- a/packages/identity/src/index.ts
+++ b/packages/identity/src/index.ts
@@ -1,8 +1,8 @@
 import type { Point } from "@zk-kit/baby-jubjub"
 import { EdDSAPoseidon, Signature, signMessage, verifySignature } from "@zk-kit/eddsa-poseidon"
 import type { BigNumberish } from "@zk-kit/utils"
-import { bufferToHexadecimal, hexadecimalToBuffer } from "@zk-kit/utils/conversions"
-import { isHexadecimal, isString } from "@zk-kit/utils/type-checks"
+import { base64ToBuffer, bufferToBase64, textToBase64 } from "@zk-kit/utils/conversions"
+import { isString } from "@zk-kit/utils/type-checks"
 import { poseidon2 } from "poseidon-lite/poseidon2"
 
 /**
@@ -82,28 +82,25 @@ export class Identity {
     }
 
     /**
-     * Returns the private key encoded as a hexadecimal string or text.
-     * @returns The private key as a hexadecimal string.
+     * Returns the private key encoded as a base64 string.
+     * @returns The private key as a base64 string.
      */
     public export(): string {
         if (isString(this._privateKey)) {
-            return Buffer.from(this._privateKey as string, "utf8").toString("hex")
+            return textToBase64(this._privateKey as string)
         }
 
-        return bufferToHexadecimal(this.privateKey as Buffer | Uint8Array)
+        return bufferToBase64(this.privateKey as Buffer | Uint8Array)
     }
 
     /**
-     * Returns a Semaphore identity based on a private key encoded as a hexadecimal string or text.
-     * @param privateKey The private key as a hexadecimal string.
+     * Returns a Semaphore identity based on a private key encoded as a base64 string.
+     * The private key will be converted to a buffer, regardless of its original type.
+     * @param privateKey The private key as a base64 string.
      * @returns The Semaphore identity.
      */
     static import(privateKey: string): Identity {
-        if (isHexadecimal(privateKey, false)) {
-            return new Identity(hexadecimalToBuffer(privateKey))
-        }
-
-        return new Identity(privateKey)
+        return new Identity(base64ToBuffer(privateKey))
     }
 
     /**
--- a/packages/identity/src/index.ts
+++ b/packages/identity/src/index.ts
@@ -12,7 +12,7 @@ import { poseidon2 } from "poseidon-lite/poseidon2"
  * and {@link https://www.poseidon-hash.info | Poseidon} for signatures.
  * In addition, the commitment, i.e. the hash of the public key, is used to represent
  * Semaphore identities in groups, adding an additional layer of privacy and security.
- * The private key of the identity can be exported as a hexadecimal string.
+ * The private key of the identity can be exported as a base64 string.
  */
 export class Identity {
     // The EdDSA private key, passed as a parameter or generated randomly.
--- a/packages/identity/src/index.ts
+++ b/packages/identity/src/index.ts
@@ -1,9 +1,8 @@
 import type { Point } from "@zk-kit/baby-jubjub"
 import { EdDSAPoseidon, Signature, signMessage, verifySignature } from "@zk-kit/eddsa-poseidon"
 import type { BigNumberish } from "@zk-kit/utils"
-import { hexadecimalToBuffer } from "@zk-kit/utils/conversions"
-import { requireString } from "@zk-kit/utils/error-handlers"
-import { isHexadecimal } from "@zk-kit/utils/type-checks"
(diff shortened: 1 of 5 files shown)


0xjei (approved):
amazing

packages/identity/src/index.ts
```diff
@@ -13,12 +12,11 @@ import { poseidon2 } from "poseidon-lite/poseidon2"
  * and {@link https://www.poseidon-hash.info | Poseidon} for signatures.
  * In addition, the commitment, i.e. the hash of the public key, is used to represent
  * Semaphore identities in groups, adding an additional layer of privacy and security.
- * The private key of the identity is stored as a hexadecimal string or text.
- * The other attributes are stored as stringified bigint.
+ * The private key of the identity can be exported as a hexadecimal string.
```
artwyman:
```suggestion
 * The private key of the identity can be exported as a base64 string.
```

packages/identity/src/index.ts
```diff
@@ -99,6 +81,28 @@ export class Identity {
...
+        if (isString(this._privateKey)) {
+            return textToBase64(this._privateKey as string)
+        }
+
+        return bufferToBase64(this.privateKey as Buffer | Uint8Array)
+    }
+
+    /**
+     * Returns a Semaphore identity based on a private key encoded as a base64 string.
+     * The private key will be converted to a buffer, regardless of its original type.
```
artwyman, line 98: This is a bit surprising, though understandable since the Base64 doesn't preserve the original type.
There needs to be a unit test to ensure this case still round-trips to produce the same identity, not just now, but also in future, since the behavior which controls that is inside of the EdDSAPoseidon constructor.

packages/identity/tests/index.test.ts
```diff
@@ -24,20 +24,59 @@ describe("Identity", () => {
...
 
+        it("Should create the same identity if the private key is the same", () => {
+            const identity = new Identity()
+
+            const identity2 = new Identity(identity.privateKey)
+
+            expect(identity.commitment).toBe(identity2.commitment)
+        })
+
         it("Should throw an error if the private key is not a string", () => {
```
artwyman, line 47:
```suggestion
        it("Should throw an error if the private key is an unexpected type", () => {
```

packages/identity/tests/index.test.ts
```diff
@@ -2,20 +2,20 @@ import { Identity } from "../src"
 
 describe("Identity", () => {
     const privateKeyText = "secret"
-    const privateKeyHexadecimal = "dd998334940df8931b76d899fdb189415f7ff4280599f03a7574725a166aad7d"
+    const privateKeyBuffer = Buffer.from(privateKeyText)
```
artwyman: I think it would be good to test a byte input which isn't the same as the string input.

packages/identity/tests/index.test.ts
```diff
@@ -24,20 +24,59 @@ describe("Identity", () => {
...
+
+            const privateKey = identity.export()
+
+            expect(typeof privateKey).toBe("string")
+            expect(Buffer.from(privateKey, "base64")).toStrictEqual(privateKeyBuffer)
+        })
+    })
+
+    describe("# import", () => {
+        it("Should import an identity", () => {
```
artwyman: As I mentioned elsewhere, I'd suggest more "round-trip" tests for export + import to ensure you end up with the same identity2 regardless of whether the original identity was constructed using default (random) key, buffer key, or text key.

packages/identity/tests/index.test.ts
```diff
@@ -2,20 +2,20 @@ import { Identity } from "../src"
...
-            expect(identity.privateKey).toHaveLength(64)
+            expect(typeof identity.privateKey).toBe("object")
+            expect(identity.privateKey).toHaveLength(32)
             expect(typeof identity.secretScalar).toBe("bigint")
             expect(identity.publicKey).toHaveLength(2)
             expect(typeof identity.commitment).toBe("bigint")
         })
 
-        it("Should create deterministic identities from a secret text (private key)", () => {
+        it("Should create a deterministic identity from a secret text (private key)", () => {
```
artwyman, line 20: These first 3 tests aren't really checking for "deterministic" output to the extent I'd expect, since they aren't checking the value of the output (just it's type and length).
I suggest having a hard-coded "expected value" for the secretScalar, commitment, and public key to check explicitly.  In this PR you'd probably generate that test data by printing the result, so it may seem like it's not testing anything new.  However it'll fail in future if the output produced by the same input ever changes, which would be a compatibility-breaking bug, and good to catch.  You could also use this approach to ensure that text vs buffer input produce the same output when they're supposed to.

artwyman (changes requested):
Functionality looks good.  I have some suggestions for additional testing.

packages/identity/src/index.ts
cedoor: So, here there are some solutions:
1. Always convert any type to buffer in the constructor.
2. Add some data to the exported private key to preserve its original type.

The first might be unexpected since a dev using text would end up with a buffer as a private key.

The second would "dirty" the exported key, which if used outside the context of the class might require some extra work when converted.

@artwyman @vplasencia I tend to prefer 2, but 1 with some extra comment could also work and keep it simple (Buffer can be easily converted to any other type). Wdyt?

artwyman: I think I personally would do 1, or nothing at all.  2 seems unnecessary if the type doesn't actually change behavior.  I.e. if buffer or string produce the same public key, then it's pretty safe to not propagate the specific TypeScript type of the object the user provided.

artwyman (approved):

vplasencia (approved):

packages/identity/src/index.ts
cedoor: Let's keep it as it is then 👍🏽

PR #810, opened 2024-06-28
perf: speed up local scripts

sripwoud:
- Execute scripts concurrently where appropriated: build, test, lint, format scripts
  - either with `yarn workspaces foreach --parallel` [option](https://yarnpkg.com/cli/workspaces/foreach#options)
  - or with [concurrently](https://github.com/open-cli-tools/concurrently)
    in which case I still group the log outputs (to avoid having them interlaced) with the `-g` option
 
 Ex of time execution reduction
 
||Before|After|
|---|---|---|
|`yarn build:libraries`|37.869s|24.732s|
|`yarn test`|22.553s|12.358s|


package.json
```diff
@@ -7,19 +7,19 @@
...
-        "build:subgraph": "yarn workspace semaphore-subgraph codegen sepolia && yarn workspace semaphore-subgraph build",
+        "build": "yarn workspaces foreach -Ap --topological-dev run build",
+        "build:libraries": "yarn workspaces foreach -Apt --no-private run build",
+        "build:subgraph": "yarn workspace semaphore-subgraph build",
         "compile:contracts": "yarn workspace semaphore-contracts compile",
-        "test": "yarn test:libraries && yarn test:contracts && yarn test:circuits && yarn test:subgraph",
+        "test": "concurrently -g -c auto \"yarn:test:*\"",
         "test:libraries": "jest --coverage",
-        "test:library": "jest packages/${0}",
+        "test.library": "jest packages/${0}",
```
sripwoud: Maybe a bit awkward, this is to avoid catching `test:library` when using the wildcard for the `test` script:
```
concurrently -g -c auto yarn:test:*
```

sripwoud: Otherwise we can explicitly list all the commands

package.json
```diff
@@ -7,19 +7,19 @@
     "bugs": "https://github.com/semaphore-protocol/semaphore/issues",
     "private": true,
     "scripts": {
-        "build": "yarn build:libraries && yarn build:subgraph",
-        "build:libraries": "yarn workspaces foreach -A -t --no-private run build",
-        "build:subgraph": "yarn workspace semaphore-subgraph codegen sepolia && yarn workspace semaphore-subgraph build",
+        "build": "yarn workspaces foreach -Ap --topological-dev run build",
```
sripwoud, line 10: This is not just a perf change.
This will run ALL the build scripts in the monorepo (including e.g the website etc...).

package.json
cedoor: I'd prefer keeping the same format for all commands with `:`, so maybe better to list all of them (they'll always be just three in the end).

cedoor (changes requested):
LGTM! Just left a comment.

cedoor (approved):

cedoor:
Ready to be merged I think @sripwoud! Thanks :)

PR #818, opened 2024-07-05
Refactor/sempahore abi to utils

### LINKED ISSUE #817: Add Semaphore ABI to utils package
**Description**

Since the [`data`](https://github.com/semaphore-protocol/semaphore/blob/main/packages/data/src/semaphoreABI.json) package and the [`subgraph`](https://github.com/semaphore-protocol/semaphore/blob/main/apps/subgraph/abis/Semaphore.json) app need the Semaphore contract interface (ABI), that JSON file must currently be copied manually every time the contract changes. The ABI could be automatically copied into the `utils` package, which can, in turn, be used by `data` and `subgraph`.

Contract interfaces are generated when the `yarn compile` command is executed in the `packages/contracts` folder. A script could be run immediately after compilation to copy the file `artifacts/contracts/Semaphore.sol/Semaphore.json` to `packages/utils/src/semaphore-abi.json`. Then all packages/apps that need the interface can import it from `@semaphore-protocol/utils`.

Additionally, all devs or external apps that need the Semaphore ABI will be able to import it directly from the `@semaphore-protocol/utils` package.

Dozie2001:
## Description

This PR adds a script that automates moving the ABI to the  ```utils``` package <br> 
Additionally, Semaphore ABI will be able to import it directly from the ```@semaphore-protocol/utils``` package

## Related Issue(s)
Closes #817 

## Checklist


package.json
```diff
@@ -10,7 +10,8 @@
         "build": "yarn workspaces foreach -Ap --topological-dev run build",
         "build:libraries": "yarn workspaces foreach -Apt --no-private run build",
         "build:subgraph": "yarn workspace semaphore-subgraph build",
-        "compile:contracts": "yarn workspace semaphore-contracts compile",
+        "copy:abi": "cp  packages/contracts/artifacts/contracts/Semaphore.sol/Semaphore.json packages/utils/src/semaphore-abi.json",
```
cedoor: `cp` will only work on Unix-like systems. A Node package alternative like [`ncp`](https://www.npmjs.com/package/ncp) should be better here.

I'd also suggest moving this script (`copy:abi`) to the `packages/contracts/package.json` file directly, otherwise if I run `yarn compile` from `packages/contracts` this script won't be executed.

cedoor: `semaphore-abi.json` actually includes bytecode and other data. `semaphore-interface.json` should be more representative.

packages/utils/src/constants.ts
```diff
@@ -1,5 +1,10 @@
+import abi from "./semaphore-abi.json"
+
 // Minimum supported tree depth.
 export const MIN_DEPTH = 1
 
 // Maximum supported tree depth.
 export const MAX_DEPTH = 32
+
+// Semaphore ABI
+export const SemaphoreABI = abi.abi
```
cedoor: Related to the comments above. Here we could export abi and bytecode. Something like:

```ts
export const SemaphoreABI = interface.abi
export const SemaphoreBytecode = interface.bytecode
```

No need to write comments if the variables are self-explanatory.

cedoor (changes requested):
Thanks @Dozie2001 :) Left some comments!

packages/utils/src/constants.ts
Dozie2001: Do you need to export Bytecode separaretly? or do you just want the option of doing so?

cedoor: What do you mean by separately? I think two exports like the ones above should be fine. Wdyt?

Dozie2001: Hmm no worrries will work on that now

cedoor:
Hey @Dozie2001, looks like you cannot use `interface` in strict mode. You should change the name of the variable.

packages/contracts/package.json
```diff
@@ -39,6 +41,7 @@
         "typescript": "^5.3.3"
     },
     "dependencies": {
-        "@zk-kit/lean-imt.sol": "2.0.0"
+        "@zk-kit/lean-imt.sol": "2.0.0",
+        "ncp": "^2.0.0"
```
cedoor: `ncp` should be a dev dependency

packages/utils/src/constants.ts
```diff
@@ -1,5 +1,10 @@
+import abi from "./semaphore-interface.json"
```
cedoor: `_interface` as a name should be ok here, as it's not the abi actually.

cedoor (changes requested):

packages/contracts/package.json
```diff
@@ -3,6 +3,7 @@
     "private": true,
     "scripts": {
         "start": "hardhat node",
+        "copy:abi": "ncp  artifacts/contracts/Semaphore.sol/Semaphore.json ../utils/src/semaphore-abi.json",
```
cedoor, line 6: This script here should be enough. No need to have one in the main `package.json` file. Just make sure to add a `&& yarn copy:abi` after `hardhat compile` here.

cedoor (changes requested):
Not related to this PR, it's an issue we're going to solve in another one. But could you remove the files in the cli-templates folder from this PR?

cedoor (approved):
LGTM @Dozie2001, thank you vm :)

PR #819, opened 2024-07-06
feat(proof): encode solidity calldata for proof

teddav:
Just a quick PR to make it easier for users to encode calldata for solidity. If using JS you probably don't need that, but I'm using Foundry, so it makes it a bit easier for me, and I thought it could be useful to others.

I also update the docs


cedoor:
Hi @teddav, the code looks good! Can you add more context on why Foundry needs the proof to be encoded?

Issue #345, opened 2023-08-15
Project Idea: Semaphore + Account Abstraction


andyguzmaneth:

jeffoodchain:
Hi @aguzmant103, is there any detail about this project? I would love to build something with this.

moven0831:
Hi @FoodChain1028, I've got some ideas about applying Semaphore on AA's wallet recovery. Since a contract-based wallet owner can assign multiple addresses as wallet guardians. We can use Semaphore to register the guardians instead of directly listing them on the smart contract. This alternation can provide privacy if the guardian's address needs to be hidden under some scenarios.

jeffoodchain:
> Hi @FoodChain1028, I've got some ideas about applying Semaphore on AA's wallet recovery. Since a contract-based wallet owner can assign multiple addresses as wallet guardians. We can use Semaphore to register the guardians instead of directly listing them on the smart contract. This alternation can provide privacy if the guardian's address needs to be hidden under some scenarios.

Hi, @moven0831. This idea looks good, yet for now we have to interact the contract-based wallet with another Externally Owned Address; thus the relationship between this address and proof in revealed as a result. A method from me to fix this is using a relayer address for everyone wants to use this service, the user would send their fee (like a transferring fee and everyone need to send the same amount of money here) and proof to the relayer and relayer would do this for the users. To be more specifically, the relayer is more like a mixer box to:
1. verify the identity in the Merkle Tree (As the owner of this AA contract).
2. mix the address to make any proof - user EOA - AA contract unrelated, and thus would preserve the privacy.

cedoor:
Hey guys, this could also be useful -> https://github.com/saleel/aa-zk-test

Author: @saleel

saleel:
Thanks for the tag @cedoor. It was an experiment I did a while back, and need to make some changes in that code to make it work correctly. 
I am also in the process of doing a write-up on explaining the whole stuff. Planning to do both in the coming weekend.

jeffoodchain:
Hi, @saleel, is there any chance or anything I can help you with?

saleel:
I wrote about it here - https://saleel.xyz/blog/zk-account-abstraction/
Here is the repo with sample code - https://github.com/saleel/semaphore-wallet

[...]

PR #875, opened 2024-10-16
Split SemaphoreVerifier key points into a separate library

### LINKED ISSUE #330: Change verification key points in contract to constant
The [`VK_POINTS` points variable](https://github.com/semaphore-protocol/semaphore/blob/main/packages/contracts/contracts/base/SemaphoreVerifier.sol#L29
) storing the verification key points in the contract is not a `constant`.

The value will be stored in storage slots and would require reading every time. If this could be changed to a `constant`, the value will be part of the bytecode and verification will be cheaper (and also better performance I think).

Since this is an array it cannot be marked as constant directly, so other techniques need to be explored.

jimmychu0807:
Split SemaphoreVerifier key points into a separate library and transform all verification key points into one big **bytes** constant.

Also added a small script that convert decimal bigint to hex string.

## Related Issue(s) 

Closes #330

Related: #345

## Other information

gas cost collected from `yarn test:report-gas`:

Original:
- Semaphore.validateProof: 315474
- Deployment SemaphoreVerifier: 10580752

With this PR:
- Semaphore.validateProof: 295635
- Deployment SemaphoreVerifier: 390597
- Deployment SemaphoreVerifierKeyPts: 3256579

@cedoor should I make a PR to semaphore `main` branch, or some other branch?

cc @saleel 

## Checklist


packages/contracts/contracts/base/SemaphoreVerifierKeyPts.sol
cedoor, line 1: What's the advantage of creating a separate lib? Couldn't this code be part of `SemaphoreVerifier.sol` directly?

cedoor (commented):
Thanks @jimmychu0807! Pointing to the main branch is ok 👍🏽

packages/contracts/contracts/base/SemaphoreVerifierKeyPts.sol
jimmychu0807: Either way is fine, no particular advantage I believe. Just that there is a big blob of bytes, so I think it is better to put it in a separate lib and make SemaphoreVerifier.sol a bit more readable.

But it is up to the core dev (you and other Semaphore colleagues) on which style to use.

cedoor: I tend to prefer having everything inside 1 contract as functions are only two in total. An alternative could also be to make the library function `internal`, and import the bytecode in `SemaphoreVerifier`, so that we can still have 1 deployment.

jimmychu0807: Updated. Yea, indeed setting the library to be internal would be better.

packages/contracts/contracts/base/SemaphoreVerifierKeyPts.sol
```diff
@@ -0,0 +1,989 @@
+// SPDX-License-Identifier: MIT
+// Part of this file was generated with [snarkJS](https://github.com/iden3/snarkjs).
+pragma solidity 0.8.23;
+
+library SemaphoreVerifierKeyPts {
+    // Verification Key points.
+    // These values are taken from the verification key json file generated with snarkjs.
+    // It allows to use the same verifier to verify proofs for all the tree depths supported by Semaphore.
+    /*
+    uint256[14][MAX_DEPTH] VK_POINTS = [
```
cedoor: I think we can remove this comment (the list of points as uints).

packages/contracts/contracts/base/SemaphoreVerifierKeyPts.sol
```diff
@@ -0,0 +1,989 @@
...
+        hex"1706e619141f6b4a93840023bb2342e759da6438e07ed974a5f1e3d506144c96"
+        hex"0bbe067b718cb861fe927716cba85cfc466eed0e9d2c1067bc7f9e356bc5d23c"
+        hex"1094541933220c2bef5ec7afe91782ec245208d58f93a08ec25eb59710a73623"
+        hex"03daf04d7a256119a10ba5f6653564dff54f8eb78577360ac477abab5ce87bf7"
+        hex"0cce5000e4686fab8ca621f1c65dda94eb8b5892b1aff79a9a5eb0a2e65efd53"
+        hex"14e0969768884cab782cafd4b81555a818559c4835fb6821ec46fd08ea418ac6"
+        hex"1ff343732becbc72450b2faedef7112b951edf2ee03bebe4b71365755e6d5518"
+        hex"10e06fbbc2176bff93433750bcbfec05aea120f193a1e4c5bf5993e098916f96";
+
+    function get_pts(uint256 merkleTreeDepth) internal pure returns (uint256[SET_SIZE] memory pts) {
```
cedoor: Method names should be camelCase.

packages/contracts/contracts/base/SemaphoreVerifierKeyPts.sol
```diff
@@ -0,0 +1,989 @@
...
+        bytes memory pt_bytes = VK_POINT_BYTES;
```
cedoor: Variable names should also be camelCase. Constants only should be snake_case (uppercase).

packages/contracts/scripts/convert-dec-to-hex.ts
cedoor, line 1: These scripts are usually part of other repos. https://github.com/vplasencia/semaphorev4-generate-zk-artifacts is what we usually use to generate the vk points in the right format.

@vplasencia should we move it there?

cedoor (changes requested):

packages/contracts/scripts/convert-dec-to-hex.ts
vplasencia: Hey! Yes, I agree, it's better to have all the scripts related to the artifacts in one place.

packages/contracts/contracts/base/SemaphoreVerifierKeyPts.sol
jimmychu0807: updated

packages/contracts/contracts/base/SemaphoreVerifierKeyPts.sol
jimmychu0807: updated

packages/contracts/contracts/base/SemaphoreVerifierKeyPts.sol
jimmychu0807: updated

packages/contracts/scripts/convert-dec-to-hex.ts
jimmychu0807: Removed the file here. 
Create an associated PR at https://github.com/vplasencia/semaphorev4-generate-zk-artifacts/pull/2

packages/contracts/contracts/base/SemaphoreVerifier.sol
```diff
@@ -23,538 +24,24 @@ contract SemaphoreVerifier {
...
-    ];
-
     // Memory data
     uint16 constant pVk = 0;
     uint16 constant pPairing = 128;
 
     uint16 constant pLastMem = 896;
 
+    constructor() {
+        SemaphoreVerifierKeyPts.checkInvariant(MAX_DEPTH);
```
jimmychu0807, line 34: @cedoor adding this invariant check at constructor to make the code more robust

cedoor: Yess, makes sense to me!

cedoor (approved):
LGTM 👍🏽

saleel (approved):
Looks good.

packages/contracts/scripts/convert-dec-to-hex.ts
saleel: I guess this should be part of the same script that generate the solidity verifier? (maybe override solidity template used by snarkjs)

PR #877, opened 2024-10-16
feat(identity): add generate commitment function

### LINKED ISSUE #873: Add function to get commitment from public key
**Is your feature request related to a problem? Please describe.**

It would be nice to create a static function in the identity package to get the commitment from a public key so that people don’t have to use the poseidon package to generate the hash. The use case can be to know which commitment is the one that goes with a signature and public key.

**Describe the solution you'd like**

```ts
static generateCommitment(publicKey: Point): bigint {
    return poseidon2(publicKey)
}
```

**Context**

This is part of the feedback from the ETHRome 2024 event.

vplasencia:
## Description

This static method is particularly useful after signature verification, as it allows retrieval of the corresponding commitment associated with the public key.

## Related Issue(s)

Closes #873

## Checklist

### DIFF
--- a/packages/identity/src/index.ts
+++ b/packages/identity/src/index.ts
@@ -138,4 +138,20 @@ export class Identity {
     static verifySignature(message: BigNumberish, signature: Signature, publicKey: Point): boolean {
         return verifySignature(message, signature, publicKey)
     }
+
+    /**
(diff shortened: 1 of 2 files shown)


cedoor (approved):
LGTM 👍🏽

PR #882, opened 2024-10-23
feat: remove prepublish script when creating template with Semaphore CLI

### LINKED ISSUE #591: Remove `prepublish` script in the CLI monorepo templates from the published package
## Description

The idea is to remove the `prepublish` script from the `scripts` object of the `package.json` file of every cli template when the template is downloaded using the CLI. 

## Context

- [`cli-template-contracts-hardhat` `prepublish` script](https://github.com/semaphore-protocol/semaphore/blob/main/packages/cli-template-contracts-hardhat/package.json#L33)

- [`cli-template-monorepo-ethers` `prepublish` script](https://github.com/semaphore-protocol/semaphore/blob/main/packages/cli-template-monorepo-ethers/package.json#L21)

- [`cli-template-monorepo-subgraph` `prepublish` script](https://github.com/semaphore-protocol/semaphore/blob/main/packages/cli-template-monorepo-subgraph/package.json#L21)

When the templates are downloaded using the [Semaphore CLI](https://github.com/semaphore-protocol/semaphore/tree/main/packages/cli), it would be nice to remove the `prepublish` script since it's not necessary for the applications that developers will create with the templates.

Before [printing the project scripts](https://github.com/semaphore-protocol/semaphore/blob/main/packages/cli/src/index.ts#L117), the `prepublish` script can be removed programmatically so that it's not part of the `package.json` of the downloaded project and it is also not part of the available scripts.

0xshikhar:
feat: to remove the prepublish script from the scripts object of the package.json file of every cli template when the template is downloaded using the Semaphore CLI.

## Related Issue(s)

Closes [Issue#591](https://github.com/semaphore-protocol/semaphore/issues/591)

## Checklist

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -34,6 +34,7 @@ const supportedTemplates = [
     }
 ]
 
+// Remove the prepublish script from the package.json file when creating a new project using the Semaphore CLI.
 function removePrePublishScript(packageJsonContent: string): string {
     try {
         const packageJson = JSON.parse(packageJsonContent)
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -13,6 +13,7 @@ import checkLatestVersion from "./checkLatestVersion.js"
 import getGroupIds from "./getGroupIds.js"
 import { getGroupId, getProjectName, getSupportedNetwork, getSupportedTemplate } from "./inquirerPrompts.js"
 import Spinner from "./spinner.js"
+import removePrePublishScript from "./removePrePublishScript.js"
 
 // Define the path to the package.json file to extract metadata for the CLI.
 const packagePath = `${dirname(fileURLToPath(import.meta.url))}/..`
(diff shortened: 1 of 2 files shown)


vplasencia (commented):
Hey @0xshikhar it looks great! Thank you. Would you mind moving the new `removePrePublishScript` function to a new file called `removePrePublishScript.ts`? Similar to the `getGroupIds.ts` file. 

Then the function will be something like `export default function removePrePublishScript(...`

0xshikhar:
Hey @vplasencia, just created a separate file for `removePrePublishScript` function to a new file called `removePrePublishScript.ts`. Thanks for your guidance.

vplasencia (commented):
Hey @0xshikhar! Great! Thank you!

Since the `updatedPackageJsonContent` variable has the correct content, maybe it's better to use it directly to list the available scripts in line [121](https://github.com/semaphore-protocol/semaphore/pull/882/files#diff-2fb448d11702324cc59d657d5945d59cd325540eb87b71f2e66a5bc01ada834cR121) instead of reading from the `package.json` file again. Then that line would be:

```ts
const { scripts } = JSON.parse(updatedPackageJsonContent)
```

instead of:

```ts
const { scripts } = JSON.parse(readFileSync(`${currentDirectory}/${projectDirectory}/package.json`, "utf8"))
```

What do you think?

0xshikhar:
Hey @vplasencia, yes I also think that will be better - not to call the function again. Just updated the PR. Thanks !

vplasencia (approved):
Great work @0xshikhar! Thank you very much! 🙏

PR #910, opened 2024-12-04
fix(group): after a sequence of updates a proof is invalid

Lauman:
## Description

When a member is removed from the group (update a leaf with 0n) and then another member is updated, the root that is saved is not correct, so the proof that is generated is not valid.
This bug is fixed by updating the lean-imt library, which was recently updated.
I detected this when I was testing the script create-mock-groups.

## Other information

The fix
https://github.com/privacy-scaling-explorations/zk-kit/pull/355

## Checklist

### DIFF
--- a/packages/contracts/test/Semaphore.ts
+++ b/packages/contracts/test/Semaphore.ts
@@ -476,24 +476,20 @@ describe("Semaphore", () => {
             const group = new Group(members)
 
             // Create a group and add 3 members.
-            let transaction = await semaphoreContract["createGroup(address)"](accountAddresses[0])
+            await semaphoreContract["createGroup(address)"](accountAddresses[0])
 
-            const { logs } = (await transaction.wait()) as any
-
-            const [groupId] = logs[0].args
+            const groupId = 2
 
             // Adding members to group
 
-            transaction = await semaphoreContract.addMembers(groupId, members)
-            await transaction.wait()
+            await semaphoreContract.addMembers(groupId, members)
 
             // Remove the third member.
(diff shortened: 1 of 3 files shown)


packages/contracts/test/Semaphore.ts
```diff
@@ -463,9 +463,66 @@ describe("Semaphore", () => {
...
+                ({ commitment }) => commitment
+            )
+            const group = new Group(members)
+
+            // Create a group and add 3 members.
+            let transaction = await semaphoreContract["createGroup(address)"](accountAddresses[0])
+
+            const { logs } = (await transaction.wait()) as any
+
+            const [groupId] = logs[0].args
```
cedoor: We know the group ID will always be 2. We could get rid of these 2 lines here.

packages/contracts/test/Semaphore.ts
```diff
@@ -463,9 +463,66 @@ describe("Semaphore", () => {
...
+            transaction = await semaphoreContract.addMembers(groupId, members)
+            await transaction.wait()
+
+            // Remove the third member.
+            {
+                group.removeMember(2)
+                const { siblings } = group.generateMerkleProof(2)
+
+                transaction = await semaphoreContract.removeMember(groupId, members[2], siblings)
+                await transaction.wait()
```
cedoor: All `await transaction.wait()` shouldn't be necessary in this tests

cedoor (changes requested):

cedoor (approved):

PR #940, opened 2025-01-18
fix: Replace map with forEach to avoid unnecessary array creation

mdqst:
## Description

`map` was used instead of `forEach` in the code. Since `map` creates a new array that isn't being used, it can be misleading and is considered an anti-pattern.

I've replaced it with `forEach` to make the intent clearer and avoid unnecessary overhead.

## Checklist

### DIFF
--- a/scripts/clean-packages.ts
+++ b/scripts/clean-packages.ts
@@ -10,7 +10,7 @@ async function main() {
         .filter((file) => file.isDirectory())
         .map((dir) => dir.name)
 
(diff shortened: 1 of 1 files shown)


vplasencia (changes requested):
Hey @mdqst! Thank you very much for this PR. 

Could you run `yarn format:write` and push the changes?

Could you update the other scripts that are in the same folder of this file following the same idea?

PR #967, opened 2025-03-09
chore: replace `map()` with `forEach()` for better clarity

0xkazak:
## Description

`map()` was being used to iterate over arrays, but the result of the map call was never used.
this is inefficient because `map()` is designed to create a new array, which is unnecessary in this case.
i’ve replaced it with `forEach()` to make the code more clear and performant, as we only need to iterate over the array and don't require the resulting array.

## Checklist

### DIFF
--- a/scripts/clean-apps.ts
+++ b/scripts/clean-apps.ts
@@ -20,7 +20,7 @@ async function main() {
         .filter((file) => file.isDirectory())
         .map((dir) => dir.name)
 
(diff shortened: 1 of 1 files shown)


vplasencia (commented):
Hey @0xkazak! Thank you very much for this PR. 

Could you update the other scripts in the same folder as this file, following the same idea?

vplasencia:
Closing this PR as the issue has been addressed in https://github.com/semaphore-protocol/semaphore/pull/1061.

PR #993, opened 2025-06-03
docs(docs): add llm setup docs

wslyvh:
## Description

Adding a guide to setup additional LLM context for development.


apps/docs/static/llms.txt
```diff
@@ -0,0 +1,1035 @@
...
+// You can also override our default zkey/wasm files.
+const proof3 = await generateProof(identity3, group, message, scope, 20, {
+    wasm: "./semaphore.wasm",
+    zkey: "./semaphore.zkey"
+})
+```
+
+----------------------------------------
+
+TITLE: Invoking Semaphore `broadcastSignal` (TypeScript)
```
ctrlc03: seems like this (and other snippets) is from v1 - not sure if it's something we should consider adding here.. as in, could that be confusing to new users? or is there a way to properly categorise versions so that appropriate suggestions are presented based on the version requested?

ctrlc03 (commented):
Thanks for this, left a comment around versioning

apps/docs/static/llms.txt
wslyvh: Thanks, you're right. Its best to only include latest version. Updated the PR!

vplasencia (approved):
Looks great @wslyvh! Thank you very much for creating this PR 🚀

PR #1006, opened 2025-07-14
Allow injecting existing PublicClient into SemaphoreViem

### LINKED ISSUE #1004: Allow Injecting an Existing Public Client into `SemaphoreViem`
### Describe the improvement you're thinking about
Currently, the `SemaphoreViem` class always [creates](https://github.com/semaphore-protocol/semaphore/blob/feb8c9c97df1fba6991963b3dbde03794597c1c4/packages/data/src/viem.ts#L134) its own PublicClient instance internally. In many modern dapp architectures (such as those using wagmi or viem context), the application already manages a PublicClient at a higher level.
It would be a significant improvement to allow users to pass an existing PublicClient via the options object.
If provided, the class should use this client instead of creating a new one. This would prevent duplicate network connections, ensure consistent configuration, and improve integration with React/wagmi-based apps.

### Describe alternatives you've considered
Continuing to require the class to always create its own client.
Downside: This leads to inefficiency, possible inconsistent state, and makes integration with frameworks like wagmi more difficult.
Using a global singleton or context to "auto-detect" a client.

### Additional context
Many dapp frameworks (wagmi, RainbowKit, etc.) encourage a single, shared PublicClient for all blockchain reads. Supporting injection will make `SemaphoreViem`` much more composable and easier to use in these environments.
This change would be backwards compatible: if no `publicClient` is provided, the class would still create its own as before.

ultraviolet10:
## Description

This PR refactors the SemaphoreViem class to allow injection of an existing PublicClient instance via the options object. If a publicClient is provided, the class will use it; otherwise, it will instantiate its own as before. This change improves composability and integration with modern dapp frameworks (such as wagmi) that already manage a PublicClient at the application level.

## Related Issue(s)

Closes: https://github.com/semaphore-protocol/semaphore/issues/1004

## Checklist

> [!IMPORTANT]
> We do not accept pull requests for minor grammatical fixes (e.g., correcting typos, rewording sentences) or for fixing broken links, unless they significantly improve clarity or functionality. These contributions, while appreciated, are not a priority for merging. If you notice any of these issues, please create a [GitHub Issue](https://github.com/semaphore-protocol/semaphore/issues/new?template=BLANK_ISSUE) to report them so they can be properly tracked and addressed.

### DIFF
--- a/packages/data/src/types/index.ts
+++ b/packages/data/src/types/index.ts
@@ -1,4 +1,4 @@
-import { Chain, Transport } from "viem"
+import { Chain, PublicClient, Transport } from "viem"
 
 export type EthersNetwork =
     | "mainnet"
@@ -65,4 +65,5 @@ export type ViemOptions = {
     transport?: Transport // Transport from viem
     chain?: Chain // Chain from viem
     apiKey?: string
+    publicClient?: PublicClient
 }
--- a/packages/data/src/viem.ts
+++ b/packages/data/src/viem.ts
@@ -131,10 +131,12 @@ export default class SemaphoreViem {
         this._options = options
 
         // Create the public client
-        this._client = createPublicClient({
-            transport,
-            chain: options.chain as Chain
-        })
+        this._client =
+            options.publicClient ??
+            createPublicClient({
+                transport,
+                chain: options.chain as Chain
+            })
 
         // Create the contract instance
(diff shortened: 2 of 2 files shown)


vplasencia (approved):
Looks great! Thank you very much 🚀

PR #1017, opened 2025-07-23
Improve decodeMessage: safer error handling and modern ethers imports

zeevick10:
## Description
- Refactored `decodeMessage` to handle invalid input gracefully (returns `null` instead of throwing).
- Updated imports to use the recommended `ethers` v6+ style.


JohnGuilding (commented):
Thanks for raising this @zeevick10! 

One issue with this change is that returning null in all cases swallows the actual errors which is useful information to the developer

Another pattern is to have `tryDothing` and `doThing` where one function fails gracefully and the other throws all errors. But not sure we want to do that here.

@vplasencia what are your thoughts on this?

PR #1025, opened 2025-08-07
fix: improve error handling in toBigInt function

Galoretka:
Fix TypeError when error.message is undefined by using optional chaining
and fallback to error.toString() for safer error handling.

### DIFF
--- a/packages/proof/src/to-bigint.ts
+++ b/packages/proof/src/to-bigint.ts
@@ -15,6 +15,6 @@ export default function toBigInt(value: BigNumberish | Uint8Array | string): big
             return _toBigInt(encodeBytes32String(value))
         }
 
-        throw TypeError(error.message)
+        throw TypeError(error?.message || error.toString())
     }
 }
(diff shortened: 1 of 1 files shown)


packages/proof/src/to-bigint.ts
```diff
@@ -15,6 +15,6 @@ export default function toBigInt(value: BigNumberish | Uint8Array | string): big
             return _toBigInt(encodeBytes32String(value))
         }
 
-        throw TypeError(error.message)
+        throw TypeError(error?.message || error.toString())
```
JohnGuilding:
```suggestion
        throw TypeError(error instanceof Error ? error.message : error.toString())
```

JohnGuilding (changes requested):
Cheers @Galoretka. One small suggestion

Galoretka:
> Cheers @Galoretka. One small suggestion

Done

vplasencia (approved):
Looks great! Thank you very much @Galoretka

PR #1028, opened 2025-08-11
docs(identity): correct @returns to bigint for secretScalar and commitment

Fibonacci747:
Fix incorrect JSDoc return type annotations in Identity getters.
Update @returns from string to bigint for secretScalar and commitment.


vplasencia (approved):
Thank you very much @Fibonacci747!

PR #1033, opened 2025-08-15
fix(data/ethers): correct MembersAdded map type to string[]

radik878:
- Change membersAddedEventsMap to Map<string, string[]> in packages/data/src/ethers.ts.
- Aligns type with actual stored values (string[]) and viem.ts logic.
- Improves type safety; no functional behavior changes.


vplasencia (approved):
Looks great! Thank you very much @radik878!

PR #1054, opened 2025-10-20
feat(data): add semaphore event listeners

### LINKED ISSUE #326: Add Semaphore event listeners
**Description**

The `@semaphore-protocol/data` library currently provides functions to download data from deployed contracts using mainly events with Ethers or TheGraph subgraphs.

However, there are no listeners, which can be useful for keeping the list of local groups, members, or proofs up-to-date.

**Resources**

* https://docs.ethers.org/v6/getting-started/#starting-contracts (Listening to events)

Lauman:
## Description

The following listeners have been added to the SemaphoreEthers class:

- onGroup (GroupCreated event): Return the ID of the newly created group.
- onMember (MemberRemoved, MemberUpdated, MemberAdded events): Return the identity confirmation of the added, updated, or deleted members. For updates , the previous identity confirmation is returned.
- onValidatedProof (ProofValidated event): Return the parameters of a validated proof.
- onGroupAdmin (GroupAdminUpdated event): Return the addresses of the old and new admins.
## Related Issue(s)
#326 

## Checklist

### DIFF
--- a/packages/data/src/ethers.ts
+++ b/packages/data/src/ethers.ts
@@ -333,4 +333,53 @@ export default class SemaphoreEthers {
         this._contract.removeAllListeners("MemberUpdated")
         this._contract.removeAllListeners("MemberRemoved")
     }
+
+    /**
+     * Listens to the ProofValidated event.
+     * @param callback Called with proof parameters.
+     */
+    onValidatedProof(
+        callback: (proof: {
+            groupId: string
+            merkleTreeDepth: number
+            merkleTreeRoot: string
+            nullifier: string
+            message: string
+            scope: string
+            points: string[]
+        }) => void
+    ): void {
+        this._contract.on(
+            "ProofValidated",
+            (groupId, merkleTreeDepth, merkleTreeRoot, nullifier, message, scope, points) => {
+                callback({
+                    groupId: groupId.toString(),
+                    merkleTreeDepth: Number(merkleTreeDepth),
+                    merkleTreeRoot: merkleTreeRoot.toString(),
+                    nullifier: nullifier.toString(),
+                    message: message.toString(),
+                    scope: scope.toString(),
+                    points: points.map((p: any) => p.toString())
+                })
+            }
+        )
+    }
+
+    offValidatedProof(): void {
+        this._contract.removeAllListeners("ProofValidated")
+    }
+
+    /**
+     * Listens to the GroupAdminUpdated event.
+     * @param callback Called with the old admin and new admin addresses.
+     */
+    onGroupAdmin(callback: (oldAdmin: string, newAdmin: string) => void): void {
+        this._contract.on("GroupAdminUpdated", (_groupId, oldAdmin, newAdmin) => {
+            callback(oldAdmin.toString(), newAdmin.toString())
+        })
+    }
+
+    offGroupAdmin(): void {
+        this._contract.removeAllListeners("GroupAdminUpdated")
+    }
 }
--- a/packages/data/src/ethers.ts
+++ b/packages/data/src/ethers.ts
@@ -412,7 +412,7 @@ export default class SemaphoreEthers {
 
     /**
      * Listens to the GroupAdminUpdated event.
-     * @param callback callback Receives the groupId, old admin and new admin addresses and event metadata.
+     * @param callback Receives the groupId, old admin and new admin addresses and event metadata.
      */
     onGroupAdminUpdated(callback: (groupId: string, oldAdmin: string, newAdmin: string, event: any) => void): void {
         this._contract.on("GroupAdminUpdated", (groupId, oldAdmin, newAdmin, event) => {
--- a/packages/data/src/ethers.ts
+++ b/packages/data/src/ethers.ts
@@ -375,7 +375,7 @@ export default class SemaphoreEthers {
 
     /**
      * Listens to the ProofValidated event.
-     * @param callback Called with proof parameters.
+     * @param callback Called with proof parameters and event metadata.
      */
     onProofValidated(
         callback: (proof: {
@@ -412,7 +412,7 @@ export default class SemaphoreEthers {
 
     /**
      * Listens to the GroupAdminUpdated event.
-     * @param callback Called with the old admin and new admin addresses.
+     * @param callback callback Receives the groupId, old admin and new admin addresses and event metadata.
      */
     onGroupAdminUpdated(callback: (groupId: string, oldAdmin: string, newAdmin: string, event: any) => void): void {
         this._contract.on("GroupAdminUpdated", (groupId, oldAdmin, newAdmin, event) => {
--- a/packages/data/src/ethers.ts
+++ b/packages/data/src/ethers.ts
@@ -295,4 +295,42 @@ export default class SemaphoreEthers {
 
         return this._contract.hasMember(groupId, member)
     }
+
+    /**
+     * Listens to the GroupCreated event.
+     * @param callback Called with the groupId of the newly created group.
+     */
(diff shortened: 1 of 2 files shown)


packages/data/src/ethers.ts
```diff
@@ -295,4 +295,91 @@ export default class SemaphoreEthers {
 
         return this._contract.hasMember(groupId, member)
     }
+
+    /**
+     * Listens to the GroupCreated event.
+     * @param callback Called with the groupId of the newly created group.
+     */
+    onGroup(callback: (groupId: string) => void): void {
```
vplasencia: What about calling this function `onGroupCreated` instead? And similarly for the others?

Then it would be `offGropupCreated` instead of `offGroup`.

packages/data/src/ethers.ts
```diff
@@ -295,4 +295,91 @@ export default class SemaphoreEthers {
...
+        this._contract.on("GroupCreated", (groupId: string) => {
```
vplasencia:
```suggestion
        this._contract.on("GroupCreated", (groupId) => {
```

We could remove the string type here.

Also, what about adding the additional `event` parameter to get additional event info? See the `Listening to Events` section here: https://docs.ethers.org/v6/getting-started/#starting-contracts

packages/data/src/ethers.ts
```diff
@@ -295,4 +295,91 @@ export default class SemaphoreEthers {
...
+
+    offGroup(): void {
+        this._contract.removeAllListeners("GroupCreated")
+    }
+
+    /**
+     * Listens to member-related events (MemberAdded, MemberRemoved, MemberUpdated).
+     * @param callback Called with member data depending on the event type.
+     */
+    onMember(callback: (eventType: string, identityCommitment: string, oldIdentityCommitment?: string) => void): void {
```
vplasencia: What do you think if we have one function per event? Something like `onMemberAdded`, `onMemberUpdated` and `onMemberRemoved`?

vplasencia (commented):
Hey @Lauman! Thank you very much for creating this PR. Great work! I added some comments.

packages/data/src/ethers.ts
Lauman: Done!

packages/data/src/ethers.ts
Lauman: Done!

packages/data/src/ethers.ts
Lauman: Fixed!

packages/data/src/ethers.ts
```diff
@@ -295,4 +295,113 @@ export default class SemaphoreEthers {
...
+    offGropupCreated(): void {
+        this._contract.removeAllListeners("GroupCreated")
+    }
+
+    /**
+     * Listens to MemberAdded events.
+     * @param callback Receives the groupId, identityCommitment and event metadata.
+     */
+    onMemberAdded(callback: (groupId: string, identityCommitment: string, event: any) => void): void {
+        this._contract.on("MemberAdded", (groupId, _index, identityCommitment, event) => {
```
vplasencia: The `MemberAdded` event is missing `merkleTreeRoot`.

See event here: https://github.com/semaphore-protocol/semaphore/blob/main/packages/contracts/contracts/interfaces/ISemaphoreGroups.sol#L31

packages/data/src/ethers.ts
```diff
@@ -295,4 +295,113 @@ export default class SemaphoreEthers {
...
+    }
+
+    /**
+     * Listens to MemberUpdated events.
+     * @param callback Receives the groupId, old identityCommitment, new identityCommitment, and event metadata.
+     */
+    onMemberUpdated(
+        callback: (groupId: string, oldIdentityCommitment: string, newIdentityCommitment: string, event: any) => void
+    ): void {
+        this._contract.on("MemberUpdated", (groupId, _index, oldIdentityCommitment, newIdentityCommitment, event) => {
```
vplasencia: The `MemberUpdated` event is missing `merkleTreeRoot`.
See event here: https://github.com/semaphore-protocol/semaphore/blob/main/packages/contracts/contracts/interfaces/ISemaphoreGroups.sol#L51

packages/data/src/ethers.ts
```diff
@@ -295,4 +295,113 @@ export default class SemaphoreEthers {
...
+    offMemberUpdated(): void {
+        this._contract.removeAllListeners("MemberUpdated")
+    }
+
+    /**
+     * Listens to MemberRemoved events.
+     * @param callback Receives the groupId, identityCommitment and event metadata.
+     */
+    onMemberRemoved(callback: (groupId: string, identityCommitment: string, event: any) => void): void {
+        this._contract.on("MemberRemoved", (groupId, _index, identityCommitment, event) => {
```
vplasencia: The `MemberRemoved` event is missing `merkleTreeRoot`.
See event here: https://github.com/semaphore-protocol/semaphore/blob/main/packages/contracts/contracts/interfaces/ISemaphoreGroups.sol#L64

packages/data/src/ethers.ts
```diff
@@ -295,4 +295,113 @@ export default class SemaphoreEthers {
...
+
+    offMemberRemoved(): void {
+        this._contract.removeAllListeners("MemberRemoved")
+    }
+
+    /**
+     * Listens to the ProofValidated event.
+     * @param callback Called with proof parameters.
+     */
+    onValidatedProof(
```
vplasencia: The event is called `ProofValidated`. I think it would be better to call this function `onProofValidated`.

vplasencia (changes requested):
Thank you for the updates. Just a few small changes and the PR is ready to be merged.

packages/data/src/ethers.ts
Lauman: I agree. Done!

packages/data/src/ethers.ts
Lauman: Thanks, fixed.

packages/data/src/ethers.ts
Lauman: Fixed!

packages/data/src/ethers.ts
Lauman: Fixed!

packages/data/src/ethers.ts
```diff
@@ -295,4 +295,132 @@ export default class SemaphoreEthers {
...
+
+    offProofValidated(): void {
+        this._contract.removeAllListeners("ProofValidated")
+    }
+
+    /**
+     * Listens to the GroupAdminUpdated event.
+     * @param callback Called with the old admin and new admin addresses.
+     */
+    onGroupAdmin(callback: (oldAdmin: string, newAdmin: string, event: any) => void): void {
```
vplasencia: Similarly, we can use `onGroupAdminUpdated` instead of `onGroupAdmin`. 

Also update parameters following the event: https://github.com/semaphore-protocol/semaphore/blob/main/packages/contracts/contracts/interfaces/ISemaphoreGroups.sol#L18

vplasencia (changes requested):

packages/data/src/ethers.ts
Lauman: You're right, fixed.

vplasencia (commented):
Would you mind adding short descriptions (as code comments before the function declarations similar to the other functions) for the functions that remove listeners? That way all the functions will have consistent documentation.

Lauman:
> short descriptions

No problem. Done.

vplasencia (approved):
Great work @Lauman! Thank you very much! 🚀

Issue #811, opened 2024-06-29
Add more info to the docs about anonymity groups.


vplasencia:

LucianoLupo:
hey! I would love to work on this one!

vplasencia:
Hey @LucianoLupo! I just assigned you this issue. Feel free to start working on it and ask any questions you may have.

PR #1062, opened 2025-10-27
Feat/add anonymity groups info

LucianoLupo:
## Description
Added some documentation about the groups

## Related Issue(s)

#811


LucianoLupo:
@vplasencia  do you think something like this makes sense ? or there was a specific aspect of the groups that wanted to be covered in #811  ?

Issue #1076, opened 2025-12-25
Constructor validation prevents reconstruction of groups with removed members


johnson86tw:
## Problem

The `Group` constructor validates that no member can be `0n`, but the `removeMember()` method sets removed members to `0n`. This creates an inconsistency where a valid group state cannot be reconstructed using the constructor with the members array.

## Steps to Reproduce

```typescript
import { Group } from "@semaphore-protocol/group"

// Create a group and remove a member
const group = new Group([1n, 2n, 3n])
group.removeMember(1)  // Sets members[1] to 0n

// Try to recreate the group from its members
const members = group.members  // [1n, 0n, 3n]
const newGroup = new Group(members)  // ❌ Throws: "Failed to add member: value cannot be 0"
```

## Current Behavior

The constructor throws an error when trying to create a group from a members array that contains `0n` values (removed members).

## Expected Behavior

The constructor should accept `0n` values.

## Impact

Users cannot reconstruct a group from its members array if any members have been removed

## Related Code
https://github.com/semaphore-protocol/semaphore/blob/1dc597a17528e521a6eef4287741e2a893c2efc5/packages/group/src/index.ts#L25-L30

vplasencia:
Hey @johnson86tw! Thank you very much for creating this issue. I will create a PR to fix it.

johnson86tw:
@vplasencia Hey, I have a [project](https://ethertw.github.io/RoyaltyAutoClaim/config) that uses Semaphore, and since it supports removing members, I’m worried that after a member is removed, admin won’t be able to reconstruct the group. Could you help prioritize fixing this issue and release a new version?

I think removing the if condition should fix the problem. Would you like me to open a PR for this?

vplasencia:
Hey @johnson86tw! Thanks for letting me know, and sorry for the delay. I will work on it today. Yes, that should solve the issue, feel free to open a PR.

johnson86tw:
@vplasencia Just opened a PR. Please help me review it when you have time. Thanks!

vplasencia:
Hey @johnson86tw! Just merged the PR and released a new version. Sorry for the delay and thank you very much!

PR #1079, opened 2026-01-19
fix(group): allow 0n values in Group constructor

johnson86tw:
re #1076


vplasencia (approved):
Looks great! Thank you very much for creating this PR @johnson86tw  🙏

PR #1081, opened 2026-01-23
fix(data): use member index from event instead of array position in S…

### LINKED ISSUE #1080: SemaphoreEthers.getGroupMembers fails when batch and single member additions are mixed
In `getGroupMembers`, the code assumes `memberAddedEvents[i]` contains the event for member at index `i`. However, `memberAddedEvents` is ordered by event occurrence, not by member index.

https://github.com/semaphore-protocol/semaphore/blob/1663372b28a36ff743a71036daceee846648122b/packages/data/src/ethers.ts#L236

This fails when members are added via both `addMembers` (batch) and `addMember` (single).

**Example:**
- Batch add at indices 0-1
- Single add at index 2

`memberAddedEvents` has 1 element, but the code tries to access `memberAddedEvents[2]` → undefined.

**Fix:** Use a map keyed by member index instead of array position.

johnson86tw:
- #1080 

@vplasencia Please help review this PR when you have time, thanks!

### DIFF
--- a/packages/data/src/ethers.ts
+++ b/packages/data/src/ethers.ts
@@ -219,6 +219,13 @@ export default class SemaphoreEthers {
 
         const memberAddedEvents = await getEvents(this._contract, "MemberAdded", [groupId], this._options.startBlock)
 
+        // Create a map from member index to identity commitment for single additions
+        const memberAddedEventsMap = new Map<string, string>()
+
+        for (const [, index, identityCommitment] of memberAddedEvents) {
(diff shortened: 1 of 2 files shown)


vplasencia (approved):
Good catch @johnson86tw!

Thank you very much for creating the issue and PR! ✨

vplasencia:
Hey @johnson86tw! I just released a new version with the fix. Thank you very much again!

PR #1094, opened 2026-09-15
fix(contracts): start the root expiration countdown when a root is replaced

### LINKED ISSUE #1093: Proofs made with the current root are rejected as soon as a member is added to a group that has been idle
**Describe the bug**

`Semaphore.sol` records the creation time of each new Merkle tree root and lets a proof use an old root while `block.timestamp <= creationDate + merkleTreeDuration`.

Because the countdown starts when the root is *created* rather than when it is *replaced*, the time a root spends as the current root is subtracted from its own grace window. If a group stays unchanged for longer than `merkleTreeDuration`, the window is already exhausted by the time the root becomes old, so proofs generated while that root was still current are rejected the moment a member is added.

The effect is the inverse of the intent: the less often a group changes, the less grace it grants. Any group whose gap between membership changes is typically longer than its duration has no working root history at all.

**To Reproduce**

1. Create a group with the default Merkle tree duration of 1 hour.
2. Add two members.
3. Let the group sit unchanged for more than 1 hour (no member added, updated or removed).
4. Generate a proof against the group's **current** root.
5. As the admin, call `addMember`.
6. Submit the proof from step 4 with `verifyProof` / `validateProof`.

It reverts with `Semaphore__MerkleTreeRootIsExpired()`, even though the root was current when the proof was generated and less than 1 hour has passed since it stopped being current.

**Expected behavior**

An old root should be accepted for `merkleTreeDuration` after it is replaced, so that proofs generated while it was the current root have time to land, regardless of how long the group was unchanged beforehand.

**Technologies**

-   Solidity version: 0.8.23
-   Package: `semaphore-contracts` v4.14.3

**Additional context**

The existing tests do not catch this: the only expiry test creates a group with a duration of `0`, which reaches expiry without needing to advance the clock, so the timing semantics are never exercised.

Related to the original root history mechanism from #98.

vplasencia:
## Description

Bug fix.

**Current behavior.** `Semaphore.sol` stamps each new Merkle tree root with its creation time and accepts an old root while `block.timestamp <= creationDate + merkleTreeDuration`. The countdown therefore starts when a root is created, not when it stops being current, and the time a root spends as the current root is consumed out of its own grace window. A group idle for longer than its duration rejects proofs made against its current root as soon as a member is added.

**New behavior.** `addMember`, `addMembers`, `updateMember` and `removeMember` now read the root *before* mutating the tree and stamp that outgoing root with the time it was replaced. The mapping is renamed `merkleRootCreationDates` -> `merkleRootSupersededDates` to match.

The current root is no longer written to the mapping at all, since `verifyProof` already matches it by equality. The number of storage writes per membership change is therefore unchanged, and the first insertion into a group now skips a cold `SSTORE` entirely.

The expiry comparison changes from `>` to `>=`. With the stamp moved to the moment of replacement, `>` would leave a duration of `0` granting one timestamp of grace; `>=` keeps the window at `[superseded, superseded + duration)` so `0` still means no old root is accepted. Callers who want to tolerate a same-timestamp race can express it with a duration of `1`.

## Related Issue(s)

Closes #1093 

## Other information

Three tests added to `packages/contracts/test/Semaphore.ts`:

-   an old root is accepted when the group was idle for longer than the duration before the update, this fails on `main` with `Semaphore__MerkleTreeRootIsExpired`
-   an old root is rejected once the duration has elapsed from the moment it was replaced
-   a root superseded by `addMembers` on a group that already has members is accepted inside its window and rejected after it

The pre-existing duration-`0` expiry test passes unmodified.

Also documented on `ISemaphore.updateGroupMerkleTreeDuration` that the duration is applied retroactively to every past root, and that it determines how long a removed member can keep proving membership with an older root.

## Checklist

> [!IMPORTANT]
> We do not accept pull requests for minor grammatical fixes (e.g., correcting typos, rewording sentences) or for fixing broken links, unless they significantly improve clarity or functionality. These contributions, while appreciated, are not a priority for merging. If you notice any of these issues, please create a [GitHub Issue](https://github.com/semaphore-protocol/semaphore/issues/new?template=BLANK_ISSUE) to report them so they can be properly tracked and addressed.


packages/contracts/contracts/Semaphore.sol
```diff
@@ -106,9 +119,12 @@ contract Semaphore is ISemaphore, SemaphoreGroups {
...
         uint256[] calldata merkleProofSiblings
     ) external override {
-        uint256 merkleTreeRoot = _removeMember(groupId, identityCommitment, merkleProofSiblings);
+        uint256 oldMerkleTreeRoot = getMerkleTreeRoot(groupId);
+
+        _removeMember(groupId, identityCommitment, merkleProofSiblings);
 
-        groups[groupId].merkleRootCreationDates[merkleTreeRoot] = block.timestamp;
+        // A group whose root is 0 has no members to remove, so the old root is never 0.
+        groups[groupId].merkleRootSupersededDates[oldMerkleTreeRoot] = block.timestamp;
```
moven0831: **Revocation now always grants a full `merkleTreeDuration` window, including for idle groups.**

Stamping the outgoing root here means a removed member's last root expires at `removalTime + merkleTreeDuration`. On `main` it expires at `rootCreationTime + merkleTreeDuration`, so a group that hadn't changed membership for longer than its duration revoked *immediately*. After this change that case gets the full window instead.

And the window isn't only "in-flight proofs get time to land" — nothing ties a root to a proof's creation time, so the removed member can mint **new** proofs (fresh scope, fresh message) against that root for the entire window. The worst case is unchanged (it was already ~`duration` right after any membership change), but the common case for a mature, low-churn group goes from instant revocation to a guaranteed `duration`-long bypass.

The NatSpec added to `ISemaphore.updateGroupMerkleTreeDuration` does mention this, but the PR description frames the change as a pure bug fix, and there's no per-root escape hatch: the only way to cut the window short is `updateGroupMerkleTreeDuration(groupId, 0)`, which simultaneously invalidates every other in-flight proof in the group and has to be manually restored.

Worth making this an explicit decision rather than a side effect. Options:

- don't stamp in `removeMember` (keeps removal instant, at the cost of dropping genuinely in-flight proofs on the removal path);
- add an admin `expireRoot(groupId, root)` so revocation doesn't require nuking the group-wide duration;
- keep as-is, but call the revocation-latency change out in the PR description / changelog so integrators sizing `merkleTreeDuration` around a revocation SLA notice it.

packages/contracts/contracts/Semaphore.sol
```diff
@@ -164,14 +180,17 @@ contract Semaphore is ISemaphore, SemaphoreGroups {
...
+            uint256 merkleRootSupersededDate = groups[groupId].merkleRootSupersededDates[proof.merkleTreeRoot];
             uint256 merkleTreeDuration = groups[groupId].merkleTreeDuration;
 
-            if (merkleRootCreationDate == 0) {
+            if (merkleRootSupersededDate == 0) {
                 revert Semaphore__MerkleTreeRootIsNotPartOfTheGroup();
             }
 
-            if (block.timestamp > merkleRootCreationDate + merkleTreeDuration) {
+            if (block.timestamp >= merkleRootSupersededDate + merkleTreeDuration) {
```
moven0831: **`merkleRootSupersededDate + merkleTreeDuration` can overflow and panic.**

`merkleTreeDuration` is unbounded — both `createGroup(address,uint256)` and `updateGroupMerkleTreeDuration` accept any `uint256` with no cap — so a large value makes this addition revert with `Panic(0x11)` rather than returning a result or one of the Semaphore custom errors. Every old-root verification in that group then fails, and stays broken until the admin lowers the duration.

The doc added in this PR ("raising it makes older roots valid again") points an integrator who wants *old roots never expire* straight at `type(uint256).max`.

Comparing by subtraction preserves the exact `[superseded, superseded + duration)` window described in the comment above and is overflow-free for any duration. `merkleRootSupersededDate` is always a `block.timestamp` from this block or an earlier one (the `== 0` check above guarantees it was written), so the subtraction can't underflow:

```solidity
if (block.timestamp - merkleRootSupersededDate >= merkleTreeDuration) {
    revert Semaphore__MerkleTreeRootIsExpired();
}
```

This is pre-existing on `main` in the `creationDate + duration` form, but the line is being rewritten here anyway and the same one-line change closes it.

moven0831: `merkleTreeDuration` isn't capped anywhere, so a very large value makes this addition overflow and revert with `Panic(0x11)` instead of a Semaphore error. The new doc above ("raising it makes older roots valid again") might nudge someone toward `type(uint256).max`.

Subtraction gives the same `[superseded, superseded + duration)` window without the overflow:

```solidity
if (block.timestamp - merkleRootSupersededDate >= merkleTreeDuration) {
```

`merkleRootSupersededDate` is always a past or current timestamp here, so it can't underflow. What do you think?

moven0831 (approved):

packages/contracts/contracts/Semaphore.sol
vplasencia: Hey @moven0831! Thank you very much for the review. The current solution should not be a problem in general since the value in production is usually not too big, but performing a subtraction achieves the same fix and is safer.

I just updated the code to use subtraction instead.

moven0831 (approved):
looks good to me!

